Issue #6, opened 2016-11-18
Write new test pattern


HaRo87:
Just for testing PRs.

gruetter:
go for it

PR #7, opened 2016-11-18
#6 fixed issue - third pattern

HaRo87:
Hi @gruetter,
here is a first test following your workflow proposal.

### DIFF
--- a/third-test.md
+++ b/third-test.md
@@ -6,7 +6,7 @@ _Third Test_
 
 ## Problem
 
-Still none.
+The Problem is that there is no problem.
 
 ## Context
 
--- a/third-test.md
+++ b/third-test.md
@@ -1,0 +1,29 @@
+# Third Test Pattern
+
+## Title
+
+_Third Test_
+
+## Problem
+
+Still none.
+
+## Context
+
+Dunno.
+
+## Forces
+
+Slight pull to the light side.
+
+## Solution
+
+Similar to _insert pattern link to [First Test](first-test.md)_
+
+## Resulting Context
+
+How would I know?
+
+## Author
+
+me


third-test.md
```diff
@@ -0,0 +1,29 @@
+# Third Test Pattern
+
+## Title
+
+_Third Test_
+
+## Problem
+
+Still none.
```
gruetter: No problem here?

third-test.md
```diff
@@ -0,0 +1,29 @@
...
+
+Dunno.
+
+## Forces
+
+Slight pull to the light side.
+
+## Solution
+
+Similar to _insert pattern link to [First Test](first-test.md)_
```
gruetter, line 21: Links only work for pattern files wich are already on `master`. At least that's what I found yesterday. Maybe I'm doing it wrong ...

gruetter (changes requested):
Hey, @HaRo87 . Let's try to make some changes ...

gruetter (approved):

third-test.md
HaRo87: I think you're doing it right. From my point of view it is OK. You will only link already existing and reviewed patterns. Otherwise you will leave a `todo` comment, or?

gruetter: yeah. Something like this.

gruetter:
Generally speaking, it seems to work as we planned. Anything missing, @HaRo87 ?

HaRo87:
Looks like it is working @gruetter. How does the stuff with the labels work when you merge?

So we need to document the workflow and publish it at innersourcecommons.org. ;-) We could do that via PR as well and invite the others to review.

gruetter:
> How does the stuff with the labels work when you merge?

Well, let's try it out. I think the labels are preserved.

gruetter:
Shall I merge?

PR #8, opened 2016-11-19
InnerSource Patterns workflow

HaRo87:
Hi @gruetter,

here comes a first draft for the InnerSource workflow description. I think it would also make sense to invite the other pattern folks as well (Tim, Cedric, Padma, ...). Will do that on Slack.

### DIFF
--- a/workflow.md
+++ b/workflow.md
@@ -29,29 +29,29 @@ contains the sources for the GitHub pages website for innersourcecommons.org.
 
 ### Working inside of the InnerSourcePatterns repository
 
-# [Create an issue] for each new idea, donut or pattern in the 
+# [Create an issue] for each new idea, donut or pattern in the
   [patterns repository][patternsRepo]. The issue should
   * contain the name of the pattern,
   * contain a short description (at least the problem) and
-  * be labeled with the appropriate label (_idea_, _donut_, _pattern_) 
-# Create a new branch either in your clone or fork of the 
+  * be labeled with the appropriate label (_idea_, _donut_, _pattern_)
+# Create a new branch either in your clone or fork of the
   [patterns repository][patternsRepo]. Please use the following pattern for
-  naming branches: `pattern/<patternName>`. Example: 
+  naming branches: `pattern/<patternName>`. Example:
   `pattern/contractedContributor`.
 # Create a _Markdown_ file with the description of the _idea_, _donut_ or
-  _pattern_ and store it either the `ideas`, `donuts` or `patterns` 
-  subdirectories. Commit and push.
-# Once your contribution is ready to be reviewed, create a pull request 
-  targeted at `master` and label it with either _idea_, _donut_ or _pattern_ 
+  _pattern_ and store it in the main directory. Commit and push.
+# Once your contribution is ready to be reviewed, create a pull request
+  targeted at `master` and label it with either _idea_, _donut_ or _pattern_
   and _Ready for Review_.
-# Reviewer can now use the PR features to comment on the pattern. 
+# Reviewer can now use the PR features to comment on the pattern.
 # After the review is complete, the reviewers should remove the label
   _Ready for Review_ and label the pattern _Accepted_ or don't apply any label
   in case rework is necessary.
 # In case of required rework, the author should apply the labels
   _Ready for Review_ and _Revised_ to indicate that a 2nd review is requested.
 # Once a pattern was labeled _Accepted_ by the reviewers, one of the TCs of the
-  [patterns repository][patternsRepo] will then merge it to `master` and 
+  [patterns repository][patternsRepo] will then move the pattern to either the
+  `ideas`, `donuts` or `patterns` subdirectories, merge it to `master` and 
   close the associated issue.
 
 ### Publishing a InnerSource pattern on innersourcecommons.org (InnerSourceCommons repository)
@@ -65,9 +65,9 @@ dedicated branch, originating from `master` starting with
 
 ## License
 
-tbd: we have to agree what license we apply to the 
-[patterns repository][patternsRepo] by default. 
- 
+tbd: we have to agree what license we apply to the
+[patterns repository][patternsRepo] by default.
+
 [commons]: http://innersourcecommons.org
 [patternsRepo]: https://github.com/paypal/InnerSourceCommons-patterns
 [commonsRepo]: https://github.com/paypal/InnerSourceCommons
--- a/workflow.md
+++ b/workflow.md
@@ -1,43 +1,54 @@
 # InnerSource Patterns Workflow
 
 This document describes the workflow for writing, reviewing and publishing
-InnerSource patterns.
-
-## General information
+InnerSource ideas, donuts and patterns. It utilizes Pull Requests to facilitate
+collaborative reviews.
 
 There are two separate repositories needed for this workflow:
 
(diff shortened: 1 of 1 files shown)


workflow.md
```diff
@@ -0,0 +1,50 @@
...
+
+This document describes the workflow for writing, reviewing and publishing
+InnerSource patterns.
+
+## General information
+
+There are two separate repositories needed for this workflow:
+
+* [InnerSourceCommons](https://github.com/paypal/InnerSourceCommons)
+* InnerSourcePatterns [todo: create and add link to repo]
```
HaRo87: The patterns repo has to be created, right?

workflow.md
```diff
@@ -0,0 +1,50 @@
...
+  * being labeled (pattern-draft, pattern, donut, ...)
+* every pattern idea, donut or draft should start on a dedicated branch,
+originating from `master` starting with `pattern/[issue-number]---[title]`
+* once a pattern idea/donut/draft is ready for review, authors should start a PR
+and label it with either `Donut`, `Idea` or `Pattern` and also with
+`Ready for Review`
+* once a pattern was accepted by the reviewers, it should be labeled `Accepted`
+and merged to `master`
+
+### Publishing a InnerSource pattern on innersourcecommons.org (InnerSourceCommons repository)
```
HaRo87, line 57: We have to define a place where we want to store the patterns on innersourcecommons.org.

workflow.md
```diff
@@ -0,0 +1,73 @@
...
+### Working inside of the InnerSourcePatterns repository
+
+# [Create an issue] for each new idea, donut or pattern in the 
+  [patterns repository][patternsRepo]. The issue should
+  * contain the name of the pattern,
+  * contain a short description (at least the problem) and
+  * be labeled with the appropriate label (_idea_, _donut_, _pattern_) 
+# Create a new branch either in your clone or fork of the 
+  [patterns repository][patternsRepo]. Please use the following pattern for
+  naming branches: `pattern/<patternName>`. Example:
```
gruetter: I have removed the issue label from the branch name, @HaRo87 as GitHub does not apply the same auto linking as Bitbucket. AFAIK, only commits can be tagged with issue ids.

workflow.md
```diff
@@ -0,0 +1,73 @@
...
+  [patterns repository][patternsRepo]. The issue should
+  * contain the name of the pattern,
+  * contain a short description (at least the problem) and
+  * be labeled with the appropriate label (_idea_, _donut_, _pattern_) 
+# Create a new branch either in your clone or fork of the 
+  [patterns repository][patternsRepo]. Please use the following pattern for
+  naming branches: `pattern/<patternName>`. Example: 
+  `pattern/contractedContributor`.
+# Create a _Markdown_ file with the description of the _idea_, _donut_ or
+  _pattern_ and store it either the `ideas`, `donuts` or `patterns`
```
gruetter: added the subdirectories as a proposal. I think it would make sense to do this in order to avoid having a user of the repo guess whether a file describes an _idea_, a _donut_ or a _pattern_.

workflow.md
```diff
@@ -0,0 +1,73 @@
...
+  in case rework is necessary.
+# In case of required rework, the author should apply the labels
+  _Ready for Review_ and _Revised_ to indicate that a 2nd review is requested.
+# Once a pattern was labeled _Accepted_ by the reviewers, one of the TCs of the
+  [patterns repository][patternsRepo] will then merge it to `master` and 
+  close the associated issue.
+
+### Publishing a InnerSource pattern on innersourcecommons.org (InnerSourceCommons repository)
+
+* for each new pattern which should be published a new issue should be created
```
gruetter, line 59: It was my understanding that we will publish the patterns on the wiki, once they are accepted in the patterns repo. Does this work with PRs, as well, @HaRo87 ?

gruetter (commented):
Done with the initial rework and review, @HaRo87 . Excellent start!

workflow.md
HaRo87: OK, makes sense to do that.

workflow.md
HaRo87: I like the idea of putting the stuff into separate directories. Having a look at the [Patterns template](https://github.com/paypal/InnerSourceCommons/wiki/InnerSource-Patterns-template) there are a couple of states a pattern can have:

> brainstormed solution (not proven), a donut (no solution), a proven solution, a reviewed pattern, an untested idea, an unreviewed draft of a proven solution

Do we need to take that into account? Like having inside of each directory a folder _proven_?

workflow.md
HaRo87: What do you mean by:

> Does this work with PRs, as well, @HaRo87 ?

I think a wiki is not the right place to publish patterns, or? In a wiki you can also edit and change stuff. On the GitHub pages site of InnerSource Commons a published pattern will have a stable (not so easy to change) state.

HaRo87 (commented):
Should we leave that PR open until some of the InnerSource Patterns folks will participate or should we close it once we're done and discuss our proposal in another way?

workflow.md
gruetter: Hmm. Here's an idea:
- On a `patterns` branch, always start the pattern `.md` file in the main directory (by convention). That way, it is obvious what file is being worked on. 
- Once the review is complete and prior to merging the `patterns` branch, it should clear what the pattern is (_idea_, _donut_ or _pattern_). So before merging, move the file to the appropriate subdirectory.

workflow.md
gruetter: > Does this work with PRs, as well, @HaRo87 ?

I was wondering whether it is possible to restrict wiki editing with PRs. I guess we can just as well store patterns as `.md` files and link to them from the wiki. I personally think a 2nd PR after the first would be a bit over the top ;)

workflow.md
HaRo87: I've added your proposal with ea6acba @gruetter.

workflow.md
HaRo87: OK, I think we have a communication issue here. ;-) I would use the GitHub pages site: innersourcecommons.org to publish the patterns, not the wiki.

HaRo87:
Should we leave that PR open until some of the InnerSource Patterns folks will participate or should we close it once we're done and discuss our proposal in another way @gruetter?

nyeates:
This PR needs to be accepted sooner than later if we are going to continue to work forward here. I assume the "Test" monicker will soon be removed from the name of this repo. There needs to be instructions like this in the root of the repo - actually they would be more visible in the README, but let's at least get a draft version out there of this. There can be additional pull requests for further changes.

NewMexicoKid:
Nick, re your comment "I assume the "Test" monicker will soon be removed from the name of this repo." -- do note that this is just a test repo; the content of this workflow will need to be copied over to the real repository that Cedric set up. 
That being said, I agree that the pull request should be accepted.

nyeates:
It might be worth considering to simply fork or replicate this repo over at InnerSourceCommons (or wherever). Otherwise, the many valuable comments, already numerous commit history (from many parties - which is a very positive thing to keep track of), and issues will be erased. Why start from zero when we have a good thing going here? Maybe there is something I don't understand though... do explain if so.

This is not the best place to put this kind of suggestion - I'll be sure to mention it elsewhere.

PR #11, opened 2016-11-30
pattern/common-requirements

nyeates:
Common code in a shared repository isn't meeting the needs of all teams that want to use it. The team that made the code available has one set of needs. Its needs are similar to what some of the receiving organization wants, but not quite the same.

This brings in the existing Pattern by Bob Hanmer, which was located here:  
https://github.com/paypal/InnerSourceCommons/wiki/Common-Requirements


nyeates:
Leaving this sit so that someone else can double check my work this time around @gruetter @HaRo87 ? 

Context: I have volunteered to help bring over some of the patterns from the wiki to github. Let me know if I am doing it right this time!

gruetter:
Looks good, @nyeates ! Thanks for kicking this off. I have added labels, assuming this pattern is ready for review.

nyeates:
Curiosity question: why does the workflow instructions have you open both an issue and a PR for new patterns? Both have the same labels and similar information. It seems redundant. Why not just have PR's?

NewMexicoKid:
For those of us with less experience on managing PRs, what is the common etiquette for merging the pull request? Who should feel responsible to check the status and to push the "Merge pull request" button? Is it a matter of availability to look at something and to make the decision? Or is there an unwritten rule on this?

Thanks for your advice.

gruetter:
>  Why does the workflow instructions have you open both an issue and a PR for new patterns?

@nyeates : I agree that the way we are using the issue tracker now is somewhat redundant. @HaRo87 and I [discussed this](https://github.com/gruetter/innerSourceTest/issues/5#issuecomment-261623470) in the issue for the workflow creation and decided to keep issues to track ideas for patterns not yet written. I guess we can omit the issue if we already have content to be added to a PR.

gruetter:
> Who should feel responsible to check the status and to push the "Merge pull request" button?

@NewMexicoKid : In the projects I know, merging is restricted to trusted committers. Thus, they are the ones who need to check the status and merge. That's why we should have more than one TC, I think.

common-requirements.md
```diff
@@ -0,0 +1,29 @@
...
+Common Requirements
+
+## Context
+Many projects are trying to use common code.  There is a shared repository that all the projects access.  This pattern applies if there is a Strong Code Owner [pattern to be written] or if there is weak code ownership, or no Benevolent Sponsor [pattern to be written]. Someone (or some project) wrote the code in the first place and contributed it to the repository. The common code is a small percentage of the overall deliverable from any of the projects.  Each project has its own delivery schedule, set of deliverables and customers.  
+Statement of Problem:  The common code in the shared repository isn’t meeting the needs of all the projects that want to use it.  
+
+## Problem Statement
+The common code in the shared repository isn’t meeting the needs of all the projects that want to use it.  
+
+## Forces
```
gruetter: Are there any relevant forces related to the human factor in all this we should consider? For me personally, InnerSource is much more about culture and the human factor than it is about code and processes.

common-requirements.md
```diff
@@ -0,0 +1,29 @@
...
+The needs of different customers are generally quite similar; however they might be expressed differently or weighted differently between customers.
+Many customers want the supplier to help them know what they need.
+The company has many “Systems Engineers” writing requirements for the products.  These requirements are supposed to be a distillation of customer needs to guide development of the product. 
+Reusing code is an important goal to save the company time and money.  
+
+## Solution
+Align the requirements of the projects so that the code that meets the requirements for one project also meets the needs for the other projects.  Decompose the code into smaller pieces for which the many using projects can agree upon requirements.  
+
+## Resulting Context
+This might require negotiating requirements changes with the customer.  It might also require other involvement by the sales teams and product managers to get alignment on the requirements.
```
gruetter: I would also mention the need to _incentivize_ reuse and to make explicitly make a decision on the components scope (which requirements are in scope, which are not).

gruetter (approved):
Always good to learn from the master. Thanks Bob.

common-requirements.md
NewMexicoKid: As a side note, one force that is human-related is that system engineers might tend to try to fill customer requirements locally within their own BL rather than reaching out to other BLs that support the same customer. Might be unrelated to this particular pattern, but this might be an InnerSource donut.

common-requirements.md
```diff
@@ -0,0 +1,29 @@
+## Title
+Common Requirements
+
+## Context
+Many projects are trying to use common code.  There is a shared repository that all the projects access.  This pattern applies if there is a Strong Code Owner [pattern to be written] or if there is weak code ownership, or no Benevolent Sponsor [pattern to be written]. Someone (or some project) wrote the code in the first place and contributed it to the repository. The common code is a small percentage of the overall deliverable from any of the projects.  Each project has its own delivery schedule, set of deliverables and customers.  
+Statement of Problem:  The common code in the shared repository isn’t meeting the needs of all the projects that want to use it.  
+
+## Problem Statement
+The common code in the shared repository isn’t meeting the needs of all the projects that want to use it.
```
NewMexicoKid: Do we need to specify in the problem statement why the common code isn't meeting the needs? I.e., ... due to conflicting requirements from the participating projects.

common-requirements.md
```diff
@@ -0,0 +1,29 @@
+## Title
+Common Requirements
+
+## Context
+Many projects are trying to use common code.  There is a shared repository that all the projects access.  This pattern applies if there is a Strong Code Owner [pattern to be written] or if there is weak code ownership, or no Benevolent Sponsor [pattern to be written]. Someone (or some project) wrote the code in the first place and contributed it to the repository. The common code is a small percentage of the overall deliverable from any of the projects.  Each project has its own delivery schedule, set of deliverables and customers.  
+Statement of Problem:  The common code in the shared repository isn’t meeting the needs of all the projects that want to use it.
```
gruetter: this line seems to be duplicated by line 9

common-requirements.md
gruetter: Just as an idea: in the forces you mention that the customers sometimes expect the company to help them clarify their requirements. My idea is that this situation should be taken advantage of by bringing about the alignment during the customer requirements negotiations and _change the customers requirements_ rather than _changing the component_.

nyeates:
It took me a bit to fully understand this. After I read it the third time, I fully got it. A future revision could work on more up-front clarification.

Once I "got it", I do like it. The decomposing of requirements into smaller pieces sounds key to me. This point could be delved into deeper.

One concern I have is that many projects do not occur at the same time, sometimes even years apart. How can requirements be co-aligned amongst teams if the former team has moved on or simply not working on it much. I guess the new team then works to become the new code owners, all whilst aligning the requirements the best they can?

common-requirements.md
```diff
@@ -0,0 +1,30 @@
...
+## Forces
+The project that made the code available has one set of needs. Its needs are similar to what some of the receiving organization wants, but not quite the same.
+Requirements on code should be derivable from real customer needs.  
+The needs of different customers are generally quite similar; however they might be expressed differently or weighted differently between customers.
+Many customers want the supplier to help them know what they need.
+The company has many “Systems Engineers” writing requirements for the products.  These requirements are supposed to be a distillation of customer needs to guide development of the product. 
+Reusing code is an important goal to save the company time and money.  
+
+## Solution
+Align the requirements of the projects so that the code that meets the requirements for one project also meets the needs for the other projects.  Decompose the code into smaller pieces for which the many using projects can agree upon requirements.
```
gruetter: I think it would help if we split this solution up into two separate bullet points: one for the requirements alignment, the other for the decomposition.

gruetter: May I suggest to use a word other than _Decompose_? Even though that's what is done (technically), the important thing to me here is that we factor out/isolate code which varies across customers and create appropriate extension points.

common-requirements.md
```diff
@@ -0,0 +1,30 @@
...
+
+Additionally, take advantage of customers expecting the supplier to help elucidate requirements. Bring about the alignment of requirements during the customer negotiations and influence the customers requirements rather than changing the component.
+
```
gruetter: From my old product line days I remember, that bringing about such an alignment requires an incentive for both the customers (usually price and TTM) _and_ for the company itself. Should we mention this here or would that unduly bloat the pattern? Maybe this could go in the _Resulting Context_ section, too.

gruetter (approved):
Excellent pattern, Robert. I wish I could be as concise as this.

common-requirements.md
NewMexicoKid: This sounds like a reasonable thing to add to the pattern at this location.

common-requirements.md
rshanmer: both done.

common-requirements.md
rshanmer: That's a good comment for resulting context.  It's a reflection that the problem (getting alignment) might introduce new problems (that require solutions like discounts).

rshanmer:
Georg's comments included and the pattern updated to the PLoP 2017 draft.

PR #13, opened 2016-12-05
pattern/contained-innersource

gruetter:
The uncontained inner sourcing of the collaboration is regarded as distracting from the needed efficiency of specific collaborative development with the other BUs sharing the use of this component, but normal development practices do not effectively work with sharing of development resources to work on this component.

/CC @nyeates , @HaRo87 , @kjstol , @psudars , @ErinMB


contained-innersource-enables-collaborative-product-development.md
```diff
@@ -0,0 +1,35 @@
+## Title  
+Contained inner source enables collaborative product development
```
gruetter: That's a mouthful ;). How about just _Contained InnerSource_?

contained-innersource-enables-collaborative-product-development.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Context
```
gruetter: Is there a previously written component of is this a green field situation? I think it matters in the case where one of the BUs is the current owner of the component.

contained-innersource-enables-collaborative-product-development.md
```diff
@@ -0,0 +1,35 @@
...
+* Multiple business units need a software component.
+* It makes sense for the business units to collaborate; each needs some different changes to the code.
```
gruetter: Maybe write that the BUs _are in agreement_ that it makes sense to collaborate. This would be more clear to me.

contained-innersource-enables-collaborative-product-development.md
```diff
@@ -0,0 +1,35 @@
...
+* One BU doesn't have the resources to do the development for all the rest.
+* InnerSource programs and tools exist.
+
+## Problem  
+The company needs to develop a software product component; multiple business units use this component.
```
gruetter: That sounds more like context to me. Since the BUs are already in agreement to collaborate and given the solution below, is the problem related to the degree of openness of a collaboration (uncontained)?

contained-innersource-enables-collaborative-product-development.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Forces  
+* The product's importance to company revenue and the committed feature content and dates require a development paradigm that provides known, stable development resources (headcount).
+
+## Sketch (optional)
+
+## Solution
+The solution is for the participating BUs to each dedicate development resources to collaboratively add the features that each BU needs to the software component. InnerSource tools and some InnerSource processes are used, but the InnerSource cooperation is contained; there is no attempt to encourage open contribution from participants outside the core group of BUs that have entered into this arrangement.
```
NewMexicoKid: Is open contribution then irrelevant to this particular pattern? I.e., it doesn't matter whether the owning BU encourages it or discourages or forbids it.
Also, how much of the open source methodology is employed in this solution to enable cooperation across BUs?

contained-innersource-enables-collaborative-product-development.md
```diff
@@ -0,0 +1,35 @@
...
+* It makes sense for the business units to collaborate; each needs some different changes to the code.
+* One BU doesn't have the resources to do the development for all the rest.
+* InnerSource programs and tools exist.
+
+## Problem  
+The company needs to develop a software product component; multiple business units use this component.
+
+## Forces  
+* The product's importance to company revenue and the committed feature content and dates require a development paradigm that provides known, stable development resources (headcount).
+
```
gruetter: Maybe add a force describing the effect of _uncontained_ InnerSource on collaborative product development. I'm guessing you are thinking about added effort for soliciting and managing contributions?

contained-innersource-enables-collaborative-product-development.md
gruetter: > Is open contribution then irrelevant to this particular pattern?

I don't think so. Not being open = not spending effort on soliciting and managing contributions outside of the group seems to be the _feature_ here. (See [my comment below](https://github.com/gruetter/innerSourceTest/pull/13#discussion_r90886211))

gruetter:
Closing the issue is a bit weird in this context. Need to better align PR and issue next time.

contained-innersource-enables-collaborative-product-development.md
NewMexicoKid: Okay.

contained-innersource-enables-collaborative-product-development.md
NewMexicoKid: Georg, this is your scenario, so is your question rhetorical? 
I had gotten the impression when we were talking about this (in the dim sum place) that it was indeed the case that one BU was the current owner of the component. Is this the case?

contained-innersource-enables-collaborative-product-development.md
NewMexicoKid: Okay.

contained-innersource-enables-collaborative-product-development.md
gruetter: I assumed this was the Nokia case. In my company, I'm not aware of any example that exactly matches this one. I guess it was a hypothetical thought.

contained-innersource-enables-collaborative-product-development.md
gruetter: Btw. @NewMexicoKid , just like in Slack, you can use Emojis to signal you _Okay_. Tends to keep the comment stream leaner.

contained-innersource-enables-collaborative-product-development.md
NewMexicoKid: Should we change it so that it matches something that you've seen work? Otherwise, this would just be a pattern idea. Did I misunderstand our discussion in the dim sum dinner? Or mis-document it?

contained-innersource-enables-collaborative-product-development.md
NewMexicoKid: We can move this to context. So would the Problem be:
The uncontained inner sourcing of the collaboration is regarded as distracting from the needed efficiency of specific collaborative development with the other BUs sharing the use of this component, but normal development practices do not effectively work with sharing of development resources to work on this component.

contained-innersource-enables-collaborative-product-development.md
NewMexicoKid: Not just the added effort but also the effect on controlling the timeline for completion and deployment.

NewMexicoKid:
Okay--I've updated the pattern to work in comments made thus far. @gruetter , could you please check that everything is okay git-wise? Also, this originally was supposed to be a pattern based on what you told us at dinner in Boston (dim sum); if it isn't, please suggest the changes needed to correct that. Thanks.

nyeates:
Has this been a successful pattern? I get that in a bigger context of lots of inner source happening at an org, that occasional baby-steps need to be taken. This shouldn't be a normal pattern though - because then you aren't really inner sourcing... your simply collaborating. However, to play devils advocate with myself, I suppose that if cross-BU code-sharing, and dev-sharing, etc is closer to open source than what would otherwise exist, it could be considered successful and in the direction of inner source.

- Could you define Contained vs Uncontained?
- The problem statement might use rephrasing.

No clear owner here, unless it was discussed elsewhere... I suggest @gruetter change this pattern as drastically as needed to make it match your experience - otherwise its dead in the water.

gruetter:
Hey @nyeates . 

> Has this been a successful pattern? I get that in a bigger context of lots of inner source happening at an org, that occasional baby-steps need to be taken. This shouldn't be a normal pattern though - because then you aren't really inner sourcing... your simply collaborating

I'd say it has been successful in two ways:
1.) It helped the organization understand the value contribution of permissionless innovation and collaboration accross business units
2.) It directly contributed to the business of Bosch.

I get your point though. The _Containment_ should go away over time. I guess we'll need a better title. How about

- Incubator
- Starting an InnerSource colony

?

> Could you define Contained vs Uncontained?

- _Contained_: The creation of InnerSource projects and communities is limited and subject to management approval. Funding is provided only for the approved projects/communities. Everybody in the company can join, contribute to and use the products of that project/community.
- _Uncontained_: No approval required to start an InnerSource project or community. Central funding is available for and distributed across all.

Does that make sense to you, @nyeates and @NewMexicoKid ?

I still consider myself the owner (sheperd?) of this pattern - hope it's not _dead in the water_ ;)

NewMexicoKid:
@gruetter -- I think that Incubator as a title, while preferable in my mind to Colony, conveys an inherent hope that it is just an incubator stage (there is some intention either among participants or those providing the infrastructure/support that it will lead to more open inner sourcing). Those actually doing the contained inner sourcing may not share that view or even fulfill that hope; and, to my mind, that's fine because there are still benefits to the pattern even without evolving to a next stage.

In terms of the definitions, it sounds as though the principle dimension for contained vs. uncontained is funding. Is that the case? I had originally thought it was more about controlling contributor access (contributions are only/mostly limited to participating BUs).

Thanks for taking ownership; is this one you will change the Owner on?

gruetter:
Meta point: I fell we're not fully leveraging the potential of PRs, yet. Ideally, the comments should be _in the code/text_ itself. If you're unsure about how to do this here's  a quick tip:
- Click on the tab named _Files changed_ above
- In the file, hover your mouse cursor at a line beginning, close to one of the `+` symbols at the beginning of the line you want to comment on. A blue `+` symbol will appear - clicking on that will bring up the dialog with which you can comment in place. 

I hope I didn't offend anyone with this ;)

gruetter:
> In terms of the definitions, it sounds as though the principle dimension for contained vs. uncontained is funding. Is that the case?
That's only a part of it. The other dimension in our case was that not everybody was _allowed_ to use our open platform from the outset. This restriction went away over time. It was IMHO a very important aspect to kick start the whole show in that it provided managers with an additional level of control.

As for the title, I agree @NewMexicoKid  that _Colony_ and _Incubator_ are probably not good choices. If anyone has a better idea, let's hear it.

gruetter:
> In terms of the definitions, it sounds as though the principle dimension for contained vs. uncontained is funding. Is that the case?

I think the principle dimension is _management control_ over what topics are tackled in InnerSource as long as it is clear what the effects of establishing InnerSource are. Funding is in my opinion only a secondary dimension. I chose the title _Contained_, as that is how I guess managers perceived our experiment.

How about the following alternative title: _Starting as an Experiment_?

gruetter:
Hey everyone. After looking at the pattern a 2nd time, I don't think that it reflects the initial situation in my company and what I wanted to convey. I'll start a new pattern in parallel, but propose to keep this one open, in case the situation described here resonates with someone. Sorry for the confusion ...

nyeates:
I do like that new name @gruetter . And I think I have heard you talk about it before. Starting grassroots, bottom up, getting buy-in by showing proof. Go forth and multiple...

ErinMB (commented):
I'm having a hard time understanding the problem statement. Can it be reworded or broken down into smaller elements to make it more clear?

PR #14, opened 2016-12-07
pattern/badly-named-piles

gruetter:
Hi everyone. Here's another pattern from the wiki for PR review.


nyeates (approved):
I like this. I was about to add a solution around having a dedicated role to sort out project metadata - and now I see it in the final bullet.

PR #15, opened 2016-12-07
Contained InnerSource - updated

NewMexicoKid:
I've made some updates to the Contained InnerSource pattern based on inputs from Georg.


gruetter:
@NewMexicoKid : you reminded me in todays pattern meeting what this pattern was all about ;). It was about a project spanning multiple business units who solved the collaboration problem using Open Source working models and infrastructure. 

The way the context is written right now is different in that it describes a company in which multiple BUs would need the *same* component and develop it collaboratively. That is IMHO a somwhat different situation and could be a pattern in its own right. What does everyone think about that?

NewMexicoKid:
Georg, I'm not sure of the differences between the case you mention and what we have in the documented pattern. It probably would be good if you could make line-edits in the pattern to make it correspond to the original situation. We can create a separate pattern idea for the other case.

gruetter:
I have rewritten the pattern and opened #84 to continue work on this, @NewMexicoKid .

PR #22, opened 2017-01-17
pattern/contracted-contributor

nyeates:
Pulled from
https://github.com/paypal/InnerSourceCommons/wiki/Contracted-Contributor

### DIFF
--- a/contracted-contributor.md
+++ b/contracted-contributor.md
@@ -1,0 +1,161 @@
+## Title
+
+_Contracted Contributor_ (aka _Free Electron_)
+
+## Problem
+
+Company A has started an InnerSource initiative. One of the goals for this
+initiative is to foster innovation by allowing every associate to voluntarily
+contribute to InnerSource projects, regardless of topic and business unit. 
+After a year, only a few voluntary contributions have been received. While
+engaging with potential individual contributors, the [[Dedicated Community Leader]]
+determined that the main problem was that the contributors respective line 
+managers did not allow them to contribute.
+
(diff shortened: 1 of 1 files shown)


contracted-contributor.md
```diff
@@ -0,0 +1,161 @@
...
+- The capacity of every associate is usually allocated to non InnerSource
+  projects for 100 %.
+- Cross organizational collaboration is not the norm.
+- Line managers usually do not have targets outside of their own organisation.
+- Contributions to InnerSource projects are expected to be made during working
+  hours, not during free time.
+
+**Review comments**
+
+* **(done)** Top level management is aware, expecting savings, putting in money. Is this budget being put into saving capacity for InnerSource? Is a barrier preventing this? Where is that money going? People don't have to hide from top level management.
```
gruetter, line 40: Same here: remove comments.

gruetter (commented):
Hey @nyeates . I apparently missed the last patterns meeting. What is the current plan? Are we going to user this PR to facilitate the 2nd review of this pattern or are we going to do this in the target repo in the PayPal instance?

contracted-contributor.md
nyeates: I didn't notice these in there. I just wholesale copied over for the time being.

We discussed in the meeting that we would like to have ongoing patterns included into master sooner. You look at the repo and it looks like nothing is going on if you dont know to dig into the ongoing PR's. I think we should initially accept pattern PR's, have something out there, and then work on modifications and comments to that base in additional PR's in the same way.

PR #25, opened 2017-01-19
Add technical steps to contribute

dicortazar:
This also fixed some numbered lists in the original file and
updated the branch naming to the current use of this.
The addition of the technical steps was required during the
last meeting and removed from the readme.md file.


dicortazar:
@kevinger I've updated the CONTRIBUTING.md file with the technical steps. I've also fixed a couple of missing numbered lists. Hope this fits with your idea in this file.

@NewMexicoKid , @gruetter I moved the technical steps from the README.md file (as they were removed for not being the place to be) to this file.

PR #38, opened 2017-02-08
pattern/get-contributions-despite-silo-thinking

gruetter:
In hierarchical and silo-organized organizations, getting voluntary contributions in InnerSource projects can be challenging. It is crucial to create mechanisms to incentivize managers to foster voluntary contributions.

Hey everyone. Here's a new donut from the Bosch guys. Have at it ;)


nyeates (approved):
Approved after editing formatting/grammar and adding an additional solution

dmittman:
First time commenter on these patterns. I've observed that although the title of this pattern seems to be "Incentive mechanisms to foster voluntary contribution" (differs from the filename, but that's a different matter), the gist of the pattern seems to be finding ways to recast the contribution effort as something other than voluntary. That is, not that the contribution is "compulsory," but that the contribution is a part of the InnerSource participant's compensation. I've done some thinking about the differences between Open Source volunteerism (where an individual is contributing to an Open Source project on their own, uncompensated time), versus InnerSource where, as an employee of a company, labor laws often prohibit you from "volunteering" your time. Is there a way to remove the word "volunteer" from InnerSource? I believe that the idea that we're "volunteering" on an InnerSource project is what sets some project managers against the idea of InnerSource.

rrrutledge:
Thanks for commenting, @dmittman .  I like that point.

spier:
We have reworked the different maturity levels of our patterns. Based on that, Donuts should be able to be merged into master much faster.

Given that this is only a Donut (i.e. no solution yet), I would suggest to rebase this branch on master, move the file to folder `patterns/1-initial` and close this PR.

Any objections @gruetter?

spier:
Trying to get this branch into a mergeable shape turned out to be hard, as it is so old.

Therefore I cherrry-picked just the Donut file, and included it in https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/254 so that it can get into the mainline soon.

Closing this PR here as we don't need it any more.

PR #43, opened 2017-02-21
pattern/review-committee

nyeatestest:
A formal review committee is setup within an org to "officiate" particular inner source projects with resources, etc. This pattern originates from Georg and Bosch.

Reviewers not yet on github:
Diogo Fregonese

### DIFF
--- a/review-committee.md
+++ b/review-committee.md
@@ -1,0 +1,44 @@
+## Title
+
+_Review Committee_ (aka _Cheese Interface_)
+
+## Context
+
+- Company A wants to introduce its first InnerSource initiative.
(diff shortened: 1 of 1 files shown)


gruetter:
Hey @NewMexicoKid : I think this pattern is ready for review.

gruetter:
Hey @nyeates , @NewMexicoKid and @ErinMB : Since this pattern is in the PLoP paper, I feel we should go forward with the review and try to finalise it before the paper is actually published.

gruetter:
Thanks for merging, @nyeates !

PR #47, opened 2017-02-22
pattern/overcoming-project-management-time-pressures

InnerSrcAdmin:
Project management believes timeline pressure and commitments on feature content does not allow for developers to spend the time needed to develop shareable code and provide support.


spier:
I found this PR because I was looking for [Donut 5: Project Management Time Pressures](https://github.com/InnerSourceCommons/innersourcecommons.org/wiki/Donut-5:-project-management-time-pressures) which is linked from the main README. That wiki page does not exist anymore and it looks like the file in this PR here was meant to replace that wiki page?

It would be great if we could get this pattern merged to `master` as is, so that others can contribute to it. 

Also if this pattern is "Proven", as the GitHub labels indicate, I would suggest to move it to the corresponding section in the main README and also adapt the status in this file.

spier:
We have reworked the different levels of our maturity patterns. Based on that, Donuts should be able to be merged into master much faster.

Given that this is only a Donut (i.e. no solution yet), I would suggest to rebase this branch on master, move the file to folder `patterns/1-initial` and close this PR.

Any objections?

PR #60, opened 2017-03-11
pattern/discover-your-innersource

nyeates:
People don't bother looking for internally developed solutions - they might not find the repo at all or be aware of its existence.


rshanmer:
Suggestion for a the first sentence in the solution:  "Make it easy to find the reusable code."
Then the other bullets support that by offering some ways to resolve the forces to achieve the above "solution".

dont-bother-looking.md
```diff
@@ -0,0 +1,49 @@
...
+    - might have poor reusability
+    - if someone put out a SW internally, the expectation is that they wouldn't have time to support it (vs. open source options)
+
+## Solution
+Make it easy to find the reusable code.
+
+* Pull in Repo names, descriptions and README.md files into the search engine
+* Implement process change to first check for internal solutions for this problem
+* Tool with a central view (but people are more inclined to google externally than look internally)
+* Concierge service (guide) to help product people find stuff. Might not scale but could be helpful in the beginning.
```
MaineC: Something I would add here is establishing a common communication channel across team boundaries. This might not scale beyond a certain organisation size. My expectation would be that people start splitting this one channel into multiple by topic once traffic gets too high.

dont-bother-looking.md
```diff
@@ -0,0 +1,49 @@
...
+* Many silos in the company; difficult to reach the developer base across those silos (a communications problem).
+* People might not want to use internal SW because they don't believe it will be helpful
+    - might not be maintained
+    - might have poor reusability
+    - if someone put out a SW internally, the expectation is that they wouldn't have time to support it (vs. open source options)
+
+## Solution
+Make it easy to find the reusable code.
+
+* Pull in Repo names, descriptions and README.md files into the search engine
```
rrrutledge: What kind of search engine are we thinking of here?

dont-bother-looking.md
```diff
@@ -0,0 +1,49 @@
...
+* Users may not know there are common places to find internally developed solutions.
+* People don't expect to find solutions internally.
+* Many silos in the company; difficult to reach the developer base across those silos (a communications problem).
+* People might not want to use internal SW because they don't believe it will be helpful
+    - might not be maintained
+    - might have poor reusability
+    - if someone put out a SW internally, the expectation is that they wouldn't have time to support it (vs. open source options)
+
+## Solution
+Make it easy to find the reusable code.
```
rrrutledge: Another avenue to explore is making it easy for the reusable code to find you.  Create ways for the people running shared internal code to discover which other projects are reasonable candidates for adoption so that they can reach out.

NewMexicoKid: Do you have some examples of what works to solve the discovery problem?

dont-bother-looking.md
NewMexicoKid: Probably whatever internal search engine people would use to try to find internally available software.

dont-bother-looking.md
rrrutledge: Search through source code at the company for projects that are candidates to use the shared internal tool and then reach out to committers to those repos.

dont-bother-looking.md
rrrutledge: That assumes that such a search engine exists.  It seems like some enterprise tooling has this kind of functionality built in, so the systems that a company is using may affect its ability to follow this part of the pattern.  In my experience systems like _GitHub_ and _npmjs_ try hard to include project search as a first-class feature.  I assume that their enterprise editions to the same, so companies using those tools would have an easier time implementing this advice.

MaineC: Even if *a* search engine exists internally, what I find really helpful is to have a one-stop-shop kind of search engine for all relevant communication and documentation. Even when using Github's enterprise offering, people often deploy additional systems like wikis to host content orthogonal to what is in the code repositories, slack channels (or IRC), mailing lists (or even nntp servers), some place to store stuff written down in office formats, search over personal e-mail etc. Several of these systems come with search built-in, but integrating this content in one search box or even just deploying a federated search engine across all sources often doesn't come off-the-shelf.

rrrutledge: Do you have examples of what a good search engine solution would be to aggregate many sources of information like that?  I've gotten the impression that it's not an easy problem.

dont-bother-looking.md
rrrutledge: * Encourage (and reward) owners of reusable code to use the same search engine to continually search for products that are candidates for use and adoption of the reusable code but not currently doing so.

PR #61, opened 2017-03-11
pattern/junkyard-styled-innersourcing

nyeates:
People don't share something so others can use it; they only share it for the sake of sharing. They wouldn't mind if no one uses it (a bit of a graveyard: "InnerSource is where modules go to die")

### DIFF
--- a/junkyard-styled-innersourcing.md
+++ b/junkyard-styled-innersourcing.md
@@ -1,5 +1,5 @@
 ## Title
-* junkyard styled inner sourcing
+* Junkyard styled inner sourcing
 * Finding but deciding not to use the internal component
 
 ## Context
@@ -9,7 +9,20 @@ Two situations:
 * Did not have the knowledge or training to create components that can b reused.
 
 ## Problem
-* People don't share something so others can use it; they only share it for the sake of sharing. They wouldn't mind if no one uses it (a bit of a graveyard: "InnerSource is where modules go to die").
+* How do you encourage people to share whatever code they write?
+
+## Forces
+* People don't share something so others can use it; they only share it for the sake of sharing. They wouldn't mind if no one uses it (a bit of a graveyard: "InnerSource is where modules go to die")
+* A developer might be a person who likes to share their code; one that likes others to read it (and maybe comment on it. But mostly they like to write code. "Hey, I wrote this nifty gzorp and you might like it (but its no skin off my nose if you don't)"
+* the code is crap
+* the code doesn't address any of the company's problems (at the moment anyway)
+* How does the person let others know that the code exists?
+* We want the person to contribute code, so what roadblocks need to be removed so that they continue making their code available?
+* How do we politely take their freely offered code and harden it or otherwise prepare it for production without having the original contributor feel like we are saying their code is crap?
+
+## Solution
+* Support all contributions to the shared repository.
+* Provide some sort of grading (Alexander marked his patterns with 0, 1 or 2 stars depending upon how convinced he was that it was a good pattern), so that the reader can be aware of what to expect. But how to do that without insulting the contributor? (solution: engage them in the process -- have 2 scores, what they think, and what users think or how many times its been used)
 
 ## Status
 Brainstormed pattern idea in progress
@@ -19,3 +32,4 @@ Brainstormed pattern idea in progress
 * Erin Bank
 * Padma Sudarsan
 * Tim Yao
+* Robert Hanmer
--- a/junkyard-styled-innersourcing.md
+++ b/junkyard-styled-innersourcing.md
@@ -1,0 +1,21 @@
+## Title
+* junkyard styled inner sourcing
+* Finding but deciding not to use the internal component
+
+## Context
+Two situations:
+
+* SW component not designed for reuse because the schedule was too tight and did not allow for the extra effort needed
+* Did not have the knowledge or training to create components that can b reused.
+
+## Problem
+* People don't share something so others can use it; they only share it for the sake of sharing. They wouldn't mind if no one uses it (a bit of a graveyard: "InnerSource is where modules go to die").
+
+## Status
+Brainstormed pattern idea in progress
+
+## Authors
+* Georg Grutter
+* Erin Bank
+* Padma Sudarsan
+* Tim Yao


rshanmer:
A bunch of thoughts:

- Since  this is a donut, it might be best to start with one of the two situations in the context and focus on it.
- When I read this problem statement I think about the person that likes to share their code -- that likes others to read it (and maybe comment on it).  But mostly they like to write code.  "Hey, I wrote this nifty gzorp and you might like it (but its no skin off my nose if you don't)"
- Is that really context and the problem is:  "How do you encourage people to share whatever code they write?"
- Then some forces might be:  * the code is crap, * the code doesn't address any of the company's problems (at the moment anyway), * How does the person let others know that the code exists?, * We want the person to contribute code, so what roadblocks need to be removed so that they continue making their code available?, * How do we politely take their freely offered code and harden it or otherwise prepare it for production without having the original contributor feel like we are saying their code is crap?
- Possible solution:  Support all contributions to the shared repository. 
- Resulting context:   Provide some sort of grading (Alexander marked his patterns with 0, 1 or 2 stars depending upon how convinced he was that it was a good pattern), so that the reader can be aware of what to expect.  But how to do that without insulting the contributor?  (soln:  engage them in the process -- have 2 scores, what they think, and what users think or how many times its been used)

junkyard-styled-innersourcing.md
```diff
@@ -0,0 +1,21 @@
+## Title
+* junkyard styled inner sourcing
+* Finding but deciding not to use the internal component
+
+## Context
+Two situations:
+
+* SW component not designed for reuse because the schedule was too tight and did not allow for the extra effort needed
+* Did not have the knowledge or training to create components that can b reused.
+
```
rshanmer, line 10: See other suggestion to pick one context.  [test]

rshanmer:
Merged before workshop at Fall Summit.

junkyard-styled-innersourcing.md
```diff
@@ -0,0 +1,35 @@
+## Title
+* Junkyard styled inner sourcing
+* Finding but deciding not to use the internal component
+
+## Context
+Two situations:
+
+* SW component not designed for reuse because the schedule was too tight and did not allow for the extra effort needed
```
gruetter, line 8: I would argue that the effort for designing SW for reuse should only be spent, if it is crystal clear that it in fact will be reused. Otherwise, YAGNI applies. So either it was clear that it would be reused and it wasn't designed that way for the reasons you mentioned, or it wasn't clear and therefore not designed for reuse (and rightly so).

junkyard-styled-innersourcing.md
```diff
@@ -0,0 +1,35 @@
...
+* Did not have the knowledge or training to create components that can b reused.
+
+## Problem
+* How do you encourage people to share whatever code they write?
```
gruetter, line 12: Hmm, the pattern title suggests to me that people are sharing, just not stuff that is ready to be reused. How about we rephrase the problem statement to s.th. like _Shared code is not reused because it was not written with reuse in mind or is not reusable._

junkyard-styled-innersourcing.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Forces
+* People don't share something so others can use it; they only share it for the sake of sharing. They wouldn't mind if no one uses it (a bit of a graveyard: "InnerSource is where modules go to die")
+* A developer might be a person who likes to share their code; one that likes others to read it (and maybe comment on it. But mostly they like to write code. "Hey, I wrote this nifty gzorp and you might like it (but its no skin off my nose if you don't)"
+* the code is crap
+* the code doesn't address any of the company's problems (at the moment anyway)
+* How does the person let others know that the code exists?
```
gruetter, line 19: I was wondering if this is a force ... as an idea: _The less effort is spent on publicising shared code the less likely it is that it will be found and reused._

junkyard-styled-innersourcing.md
```diff
@@ -0,0 +1,35 @@
...
+## Context
+Two situations:
+
+* SW component not designed for reuse because the schedule was too tight and did not allow for the extra effort needed
+* Did not have the knowledge or training to create components that can b reused.
+
+## Problem
+* How do you encourage people to share whatever code they write?
+
+## Forces
```
gruetter, line 14: @rshanmer : Some of the forces in this pattern are formulated as questions, rather than e. g. tradeoffs. Is this a valid way to phrase forces?

junkyard-styled-innersourcing.md
```diff
@@ -0,0 +1,35 @@
...
+* People don't share something so others can use it; they only share it for the sake of sharing. They wouldn't mind if no one uses it (a bit of a graveyard: "InnerSource is where modules go to die")
+* A developer might be a person who likes to share their code; one that likes others to read it (and maybe comment on it. But mostly they like to write code. "Hey, I wrote this nifty gzorp and you might like it (but its no skin off my nose if you don't)"
+* the code is crap
+* the code doesn't address any of the company's problems (at the moment anyway)
+* How does the person let others know that the code exists?
+* We want the person to contribute code, so what roadblocks need to be removed so that they continue making their code available?
+* How do we politely take their freely offered code and harden it or otherwise prepare it for production without having the original contributor feel like we are saying their code is crap?
+
+## Solution
+* Support all contributions to the shared repository.
```
gruetter, line 24: I like the idea behind this but I feel this needs to be fleshed out a little. Here are some ideas:
- Support the users in writing proper `readme.md` and `contribute.md` files, e. g. by providing templates
- Announce new shared repositories in a weekly/monthly blog post and give the authors a chance to make themselves and their SW known to others

junkyard-styled-innersourcing.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Forces
+* People don't share something so others can use it; they only share it for the sake of sharing. They wouldn't mind if no one uses it (a bit of a graveyard: "InnerSource is where modules go to die")
+* A developer might be a person who likes to share their code; one that likes others to read it (and maybe comment on it. But mostly they like to write code. "Hey, I wrote this nifty gzorp and you might like it (but its no skin off my nose if you don't)"
+* the code is crap
+* the code doesn't address any of the company's problems (at the moment anyway)
+* How does the person let others know that the code exists?
+* We want the person to contribute code, so what roadblocks need to be removed so that they continue making their code available?
+* How do we politely take their freely offered code and harden it or otherwise prepare it for production without having the original contributor feel like we are saying their code is crap?
+
```
gruetter, line 22: Suggestion for additional force: _The more responsive the original author is to requests for changes coming from potential re-users of her code, the more likely it is that it will be reused._

PR #64, opened 2017-03-16
pattern/not-invented-here

nyeates:
Perfectly good solutions are being rejected because of Not Invented Here. Engineers and their managers will choose to implement functionality themselves first, even if an alternative exists.


not-invented-here.md
```diff
@@ -0,0 +1,108 @@
...
+Finding incentives to drive IS behavior (incentives and measures can vary at a team level)
+Shift to a "profoundly found elsewhere" culture
+Demonstrate organizational interest in outside opinions
+Identify influencers who agree to be early adopters - set the stage for others
+Contributability is a mark of quality
+
+
+## Desired Resulting Context
+What is the situation after the problem has been solved?
+Developers search for and leverage existing options as opposed to rewriting. This results in: increased efficiency, increased reuse, higher levels of developer satisfaction, increased speed to market.
```
rrrutledge: Developers interact with code and products produced by others with the same trust and engagement as those that they themselves have produced.

spier: I have worked this feedback. Thanks!

spier:
This pattern seems to have been created at one of the early ISC Summits as well, as it contains many notes from conversations that seem to have happened in-person.

To get this pattern into a mergeable state, one would have to remove those notes and rework the content such that it becomes understandable for anybody interested in the Not-Invented-Here Mindset in the context of InnerSource.

patterns/1-initial/not-invented-here.md
```diff
@@ -0,0 +1,136 @@
+## Title
+
+Overcoming the Not-Invented-Here Mindset
+
+## Patlet
+
+TBD
```
spier:
```suggestion
Perfectly good solutions are being rejected because of "Not Invented Here" (NIH). Engineers and their managers will choose to implement functionality themselves first, even if an alternative exists. A shift towards a culture of "Proudly Found Elsewhere" can help reduce the negative impact of NIH.
```

Add draft Patlet based on information in the PR and a bit from the Solution section.

spier:
I changed my mind on this PR.

The pattern is certainly in an early draft state e.g. it contains many notes that might have been written at an ISC Summit where this draft was ideated.

However I still think that this draft would create more value if we merge it into main, as that makes it more likely for readers to find it.

README.md
```diff
@@ -119,7 +120,12 @@ The pattern form is useful for describing proven patterns but it can also be use
 
 We welcome your contribution - be it small or huge! To learn more about how you can become a contributor, please see our [CONTRIBUTING.md](CONTRIBUTING.md) file.
 
-## Licensing
+
+## Related References
+
+* [A pattern language for pattern writing](http://hillside.net/index.php/a-pattern-language-for-pattern-writing), Meszaros and Doble
+
+
```
```suggestion

## Licensing

```

spier (approved):
Approving this PR on the premise that:
a) the PR is so old that it is unlikely that the original authors will still work on it and
b) it is better to have a half-finished pattern draft in the main branch, than in a PR, as readers are more likely to find it in the main branch

PR #65, opened 2017-03-21
Add webinar link to README

nyeates:
We had our webinar and wanted to include it on the front README page. So, I redid the first few links to be more readable/digestable.


NewMexicoKid (approved):
Looks great--thanks for doing this, Nick.

PR #66, opened 2017-03-22
Added new pattern: Start as Experiment

gruetter:
Hi patterns community. Here's a new pattern which describes a solution which we have (I think) successfully applied at Bosch. I'm looking forward to your feedback and suggestions.

Cheers,
Georg

### DIFF
--- a/start-as-experiment.md
+++ b/start-as-experiment.md
@@ -2,6 +2,9 @@
 
 Start as an Experiment
 
+# Patlet
+
+
 # Problem
 
 An InnerSource initiative is considered but not started because management is 
@@ -10,19 +13,18 @@ investment.
 
 # Context
 
-The company is considering to leverage InnerSource to make the required
-collaboration more efficient. However, most managers are not familiar with the
-Open Source working model and are instead accustomed to hierarchical, top-down
-control style management. The idea of InnerSource is very popular with software
-developers in the company, not the least because many developers use or are
-actively developing Open Source software.
+The company is considering to leverage InnerSource to increase the efficiency
+of collaboration on software projects. However, most managers are not familiar
+with the Open Source working model and are instead accustomed to hierarchical,
+top-down control style management. The idea of InnerSource is very popular with
+software developers in the company, not the least because many developers use
+or are actively developing Open Source software.
 
 # Forces
 
 - Managers will want to validate the claims of improved collaboration through
   InnerSource before making a long term investment. This usually involves
-  putting numbers against the expected improvements, which are hard to come
-  by for InnerSource.
+  measuring the improvements.
 - If the InnerSource initiative will likely have a huge uptake among developers
   and if many projects are likely to rely on it, a decision to shut it down
   will be very unpopular and therefore hard to make. The perceived resulting
@@ -55,18 +57,19 @@ Consider designating the end of the experiment a _pivot_, _change_ or _pause_
 point to re-evaluate. Also consider establishing a [Review
 Committee](review-committee.md) to increase the chances of management buy-in
 through participation. Depending on company culture, it might be helpful to
-accompany the experiment with appropriate metrics (**tbd: Pattern link**). If
-the projects in the experiment don't provide a direct impact on the companies
-revenue, consider introducing [Cross Team
-Valuation](https://github.com/paypal/InnerSourcePatterns/blob/rrrutledge-patch-3/crossteam-project-valuation.md)
-to highlight its value contribution.
+accompany the experiment with appropriate metrics [First Steps With
+Metrics](introducing-metrics-in-innersource.md). If the projects in the
+experiment don't provide a direct impact on the companies revenue, consider
+introducing [Cross Team Valuation](crossteam-project-valuation.md) to highlight
+their value contributions.
 
 # Resulting Context
 
 Managers are able to kick start InnerSource for the following reasons:
 
-- The experimental setup eases the need to management to scrutinize their
-  numbers in the same way that they would for typical projects.
+- The experimental setup eases the need of managers to scrutinize the
+  InnerSource program numbers in the same way that they would for typical
+  projects.
 - The possibility of failure of the experiment is understood and accepted. The
   personal risk for the supporting managers is minimized.
 - Even in case of a failure, the setup ensures that the company will learn from
@@ -74,13 +77,14 @@ Managers are able to kick start InnerSource for the following reasons:
 - In case of success, the data gathered during the experiment will allow
   managers to make a longer lasting commitment to InnerSource. 
 
-Participants in the InnerSource experiment are conscious of the fact that they
-have to prove to management that InnerSource yields the promised benefits. It
-will therefore help to focus work on those activites which provide the most and
-most demonstrable value thus increasing the chances of success.
+Participants in the InnerSource experiment are now conscious of the fact that
+they have to prove to management that InnerSource yields the promised benefits.
+It will therefore help to focus work on those activites which provide the most
+demonstrable value thus increasing the chances of success.
 
-Finally, starting as an experiment makes it much easier to bend and adapt the 
-existing processes and rules before you break them.
+Finally, starting as an experiment makes it much easier to sidestep regulations
+and forces such as tool and process policies which could decrease the chances
+of success.
 
 # Related Patterns
 
@@ -89,8 +93,7 @@ existing processes and rules before you break them.
 
 # Known Instances
 
-- Robert Bosch GmbH
-
+- Robert Bosch GmbH (globally distributed software development)
 
 # Author
 
@@ -98,7 +101,7 @@ existing processes and rules before you break them.
 
 # Status
 
-Draft Pattern
+Proven Pattern
 
 # Acknowledgements
 
@@ -106,3 +109,6 @@ Draft Pattern
 - Diogo Fregonese (Robert Bosch GmbH)
 - Robert Hansel (Robert Bosch GmbH)
 - Hans Malte Kern (Robert Bosch GmbH)
+- Russ Rutledge (Nike)
+- Tim Yao (Nokia)
+- Clint Cain (Optum)
--- a/start-as-experiment.md
+++ b/start-as-experiment.md
@@ -1,0 +1,81 @@
+# Title
+
+Start as an Experiment
+
+# Problem
+
+An InnerSource initiative is considered but not started because management is 
+unsure about its outcome and, as a result, is not willing to commit to an 
+investment.
+
+# Context
+
+The majority of software development projects in the company are distributed
+globally. The company is considering to leverage InnerSource to make the
+required collaboration more efficient. However, most managers are not familiar
+with the Open Source working model and are instead accustomed to hierarchical,
+top-down control style management. The idea of InnerSource is very popular with
+software developers in the company, not the least because many developers use
+or are actively developing Open Source software.
+
+# Forces
+
+- Managers will want to validate the claims of improved collaboration through
+  InnerSource before making a long term investment. This usually involves
+  putting numbers against the expected improvements, which are hard to come
+  by for InnerSource.
+- If the InnerSource initiative has a huge uptake among developers and if many
+  projects come to rely on it, a decision to shut it down will be very 
+  unpopular and therefore hard to make.
+- When implementing new working models such as InnerSource which are radically
+  different from working models previously practiced in the company, it is 
+  likely that existing, mandatory processes are no longer applicable and even
+  that one has to operate in a regulatory, sometimes legal no-mans land.
+
+# Solution
+
+Declare the InnerSource initiative as a time limited experiment. Define and
+communicate the criteria for evaluating the experiment and ensure that 
+sufficient data is gathered to perform the evaluation. Consider establishing
+a _Review Committee_ (**tbd**: pattern link) to increase the chances of 
+management buy-in.
+
+# Resulting Context
+
+Managers are able to kick start InnerSource for the following reasons:
+
+- The experimental setup eases the need to management to scrutinize their
+  numbers in the same way that they would for typical projects.
+- The possibility of failure of the experiment is understood and accepted. The
+  personal risk for the supporting managers is minimized.
+- Even in case of a failure, the setup ensures that the company will learn from
+  the experiment.
+- In case of success, the data gathered during the experiment will allow
+  managers to make a longer lasting commitment to InnerSource. 
+
+Participants in the InnerSource experiment are conscious of the fact that, as
+an experiment, the InnerSource initiative could be shut down or not continued
+after the planned lifetime in case the experiment does not yield the expected 
+results.
+
+Finally, starting as an experiment makes it much easier to bend and adapt the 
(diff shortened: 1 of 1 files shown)


start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Forces
+
+- Managers will want to validate the claims of improved collaboration through
+  InnerSource before making a long term investment. This usually involves
+  putting numbers against the expected improvements, which are hard to come
+  by for InnerSource.
+- If the InnerSource initiative has a huge uptake among developers and if many
+  projects come to rely on it, a decision to shut it down will be very 
+  unpopular and therefore hard to make.
```
NewMexicoKid: If many projects come to rely on InnerSource, doesn't that mean that it is successful? Would a decision to shut it down then be very unlikely? Does this force represent a fear that management has about even starting this (a worry that they would lose control over the program because of its popularity)?

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
...
+- When implementing new working models such as InnerSource which are radically
+  different from working models previously practiced in the company, it is 
+  likely that existing, mandatory processes are no longer applicable and even
+  that one has to operate in a regulatory, sometimes legal no-mans land.
```
NewMexicoKid: Can you elaborate on what some of the regulatory and legal concerns are?

start-as-experiment.md
gruetter: > If many projects come to rely on InnerSource, doesn't that mean that it is successful?

That would be my definition of success, yes. But then, I'm not a manager. The benefits of InnerSource can sometimes be hard to quantify, especially the soft factors, such as employee satisfaction, improved learning and increased networking. Not being able to put hard numbers on e. g. the increase in efficiency might make InnerSource look like a failure to some managers.

> Would a decision to shut it down then be very unlikely?

I think there are shades of grey, here. If the company does not support InnerSource on the highest level anymore (e. g. no funding for central platform), the projects which are relying on InnerSource would most likely maintain a more islolated version of it for their particular use. I guess you could call that InnerSource, too, but it would probably be a step back from the values of _openness_ and _transparency_. 

> Does this force represent a fear that management has about even starting this (a worry that they would lose control over the program because of its popularity)?

Yes!

start-as-experiment.md
gruetter: I know of three areas, which were (are) problematic, at least here in Europe:
- Tax problems (a project in one country contributing SW for free, a project in another country makes money of of it)
- Open Source licensing obligations: In my company, there are strict processes ensuring that SW that we publish and that uses OSS complies with the requirements spelled out in their respective licenses. Technically, pushing SW to a platform that is accessible by multiple legal entities is already an act of publishing. I'm guessing that most companies which haven't practiced InnerSource do not have regulations for that in place.
- Export control regulations: There are a lot of export control regulations (most of them put in place by the US) that apply to SW being published (same case as above).

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Solution
+
+Declare the InnerSource initiative as a time limited experiment. Define and
```
gruetter: from Erin: mention example for concrete criteria for measuring success

gruetter: From what I was able to gather so far, (this was before I engaged with InnerSource at Bosch), the criteria put in place were used to evaluate the _health_ of the InnerSource communities and not to evaluate the benefits of _InnerSource_. The thinking was that we need to assess first whether or not InnerSource can flourish within the organisation and that introducing "hard" measurements (efficiency KPIs and such) would substantially alter the nature and outcome of the experiment. Here are some of the criteria put in place for our experiment:

- Geographical distribution (of developers)
- Departmental mix (of developers)
- Openness of communication (within community)
- Career path based on merit (within community)
- Democratic decisions (within community)

I have to admit that this is somewhat different from how I wrote it in this pattern.  @ErinMB, @NewMexicoKid and @nyeates: should we go about this?

gruetter: My opinion: I would rewrite this part of the solution section to better reflect how we approached it at Bosch.

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
...
+globally. The company is considering to leverage InnerSource to make the
+required collaboration more efficient. However, most managers are not familiar
+with the Open Source working model and are instead accustomed to hierarchical,
+top-down control style management. The idea of InnerSource is very popular with
+software developers in the company, not the least because many developers use
+or are actively developing Open Source software.
+
+# Forces
+
+- Managers will want to validate the claims of improved collaboration through
```
nyeates, line 25: From 7/18/17 patterns meeting: Engineering-led inner source often does not scale to the wider organization because engineers lack a marketing skillset and upward/sideways org traversal.

Are we able to weave this interesting comment/observation in anywhere?

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
...
+- The experimental setup eases the need to management to scrutinize their
+  numbers in the same way that they would for typical projects.
+- The possibility of failure of the experiment is understood and accepted. The
+  personal risk for the supporting managers is minimized.
+- Even in case of a failure, the setup ensures that the company will learn from
+  the experiment.
+- In case of success, the data gathered during the experiment will allow
+  managers to make a longer lasting commitment to InnerSource. 
+
+Participants in the InnerSource experiment are conscious of the fact that, as
```
nyeates: Comment from 7/18/17 patterns meeting: Consider having a "pivot, change, pause" point. Instead of an end-date, you re-evaluate

start-as-experiment.md
rrrutledge: I'm confused, though.  These concerns wouldn't apply to InnerSource efforts (within a single company), right?

NewMexicoKid: @rrrutledge , I think that the tax concerns would still apply within a single company if that company spanned several countries (each with different tax laws); but I am not a tax expert.

gruetter: @NewMexicoKid is right. Bosch, e. g., has hundreds (!) of legal entities in more than 100 countries world wide.

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Problem
+
+An InnerSource initiative is considered but not started because management is 
+unsure about its outcome and, as a result, is not willing to commit to an 
+investment.
+
+# Context
+
+The majority of software development projects in the company are distributed
```
rrrutledge: I'm not sure if the global distribution of developers is relevant context for this pattern.  It feels like the following sections would all still apply even without this context.

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
...
+  different from working models previously practiced in the company, it is 
+  likely that existing, mandatory processes are no longer applicable and even
+  that one has to operate in a regulatory, sometimes legal no-mans land.
+
+# Solution
+
+Declare the InnerSource initiative as a time limited experiment. Define and
+communicate the criteria for evaluating the experiment and ensure that 
+sufficient data is gathered to perform the evaluation. Consider establishing
+a _Review Committee_ (**tbd**: pattern link) to increase the chances of
```
rrrutledge: Link to [review committee](https://github.com/paypal/InnerSourcePatterns/blob/master/review-committee.md)?

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
...
+# Forces
+
+- Managers will want to validate the claims of improved collaboration through
+  InnerSource before making a long term investment. This usually involves
+  putting numbers against the expected improvements, which are hard to come
+  by for InnerSource.
+- If the InnerSource initiative has a huge uptake among developers and if many
+  projects come to rely on it, a decision to shut it down will be very 
+  unpopular and therefore hard to make.
+- When implementing new working models such as InnerSource which are radically
```
rrrutledge: Split this bullet into 2-3 sentences.

start-as-experiment.md
rrrutledge: I don't understand these last two points in the resulting context - how are they good, desirable, or related to the problem or opening context?

start-as-experiment.md
rrrutledge: Got it.

start-as-experiment.md
rrrutledge: Is this comment/observation directly related to the thrust and intent of this pattern?

start-as-experiment.md
rrrutledge: Sounds good to me.  Would the right thing to do be to reference another pattern.  I see two in pull request that may be related ([Cross-team Project Valuation](https://github.com/paypal/InnerSourcePatterns/pull/87), [Add pattern for discussion: first steps with metrics](https://github.com/paypal/InnerSourcePatterns/pull/83)).  Or create a new one yourself?

start-as-experiment.md
rrrutledge: Sounds good.

rrrutledge (changes requested):

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
+# Title
+
+Start as an Experiment
+
```
rrrutledge, line 4: Interested in a **Patlet** in this pattern?

start-as-experiment.md
NewMexicoKid: It's a condition of a (the) known case; I think it doesn't do any harm to mention this point, even if the pattern applies without it. (perhaps there could be a parenthetical statement about how the global distribution is not believed to be a necessary condition for applying this pattern?)

start-as-experiment.md
gruetter: I think this is not directly related, @nyeates. This force was geared towards the desire of mgmt. to validate a new idea before they sign off on it. So it's more related to KPIs and measurements than marketing skills.

start-as-experiment.md
gruetter: I like it, too!

gruetter: Hmm, you have a point there, @rrrutledge . What I wanted to express is that being transparent about the fact that the experiment might not continue would help avoid unnecessary frustration in the developer base. I can say that this wouldn't have helped me personally, though. One way in which the experimental nature of our InnerSource initiative did help was that it communicated to us developers that we had to prove that InnerSource (which we all love) does work well, indeed. Does that resonate with you, @rrrutledge ?

start-as-experiment.md
gruetter: absolutely!

start-as-experiment.md
```diff
@@ -0,0 +1,81 @@
+# Title
+
+Start as an Experiment
```
rrrutledge, line 3: I think that this pattern may be rediscovering a lot of **Trial Run** from the [_Fearless Change Pattens_](http://www.fearlesschangepatterns.com/).  It's awesome because it means that the content and idea here is solid, but I think it would be good (in the long run) to think about how patterns here relate to what is already documented and printed there.

start-as-experiment.md
gruetter: I have added text to explain how managers might fear loss of control due to huge uptake. Not sure how to factor in the rest of your feedback, though, @NewMexicoKid . Any suggestions?

start-as-experiment.md
gruetter: I've updated the force to reflect your feedback.

start-as-experiment.md
gruetter: I've added your idea to replace the _end point_ with a _pivot/pause/change point_, @nyeates .

start-as-experiment.md
gruetter: I've enhanced the solution and linked to the (not yet existing) patterns as you suggested, @rrrutledge .

start-as-experiment.md
gruetter: The way I have rewritten this hopefully makes it clear that global distribution is just one example for a criteria.

start-as-experiment.md
gruetter: done

start-as-experiment.md
rrrutledge: I think that parenthetical statement would be good.  Wouldn't want anyone to be discouraged from trying out this pattern because they don't believe it applies to them.

start-as-experiment.md
gruetter: done. It was a very german sentence, indeed ;)

start-as-experiment.md
gruetter: I've rewritten this paragraph. Does this work better for you, @rrrutledge ?

start-as-experiment.md
gruetter: The connection is obvious, indeed. Are you suggesting to add a section _Related Work_ or something similar?

start-as-experiment.md
```diff
@@ -0,0 +1,102 @@
...
+- Democratic decision making within community.
+
+Consider designating the end of the experiment a _pivot_, _change_ or _pause_
+point to re-evaluate. Also consider establishing a [Review
+Committee](review-committee.md) to increase the chances of management buy-in
+through participation. Depending on company culture, it might be helpful to
+accompany the experiment with appropriate metrics (**tbd: Pattern link**). If
+the projects in the experiment don't provide a direct impact on the companies
+revenue, consider introducing [Cross Team
+Valuation](crossteam-project-valuation.md) to highlight its value contribution.
```
rrrutledge: Thanks for thinking of this document!  It is not in `master` yet so you'll have to link to [the branch](https://github.com/paypal/InnerSourcePatterns/blob/rrrutledge-patch-3/crossteam-project-valuation.md) for now if you want to reference it.

start-as-experiment.md
rrrutledge: I think that would be good for now.  In the long run I'm wondering if we explicitly just start referring people to go read patterns from that book and just focus on any innersource-specific adaptations/examples in our documentation.

start-as-experiment.md
```diff
@@ -0,0 +1,102 @@
...
+- The experimental setup eases the need to management to scrutinize their
+  numbers in the same way that they would for typical projects.
+- The possibility of failure of the experiment is understood and accepted. The
+  personal risk for the supporting managers is minimized.
+- Even in case of a failure, the setup ensures that the company will learn from
+  the experiment.
+- In case of success, the data gathered during the experiment will allow
+  managers to make a longer lasting commitment to InnerSource. 
+
+Participants in the InnerSource experiment are conscious of the fact that they
```
rrrutledge: > I've rewritten this paragraph. Does this work better for you, @rrrutledge

Yes, this paragraph makes sense to me.  Still not able to connect the dots on the one below, though.

rrrutledge (commented):
Thanks for these updates.

start-as-experiment.md
gruetter: I agree with Russ that the pattern will apply regardless of whether or not SW development is globally distributed or not. I'll remove it. Thanks for persisting, @rrrutledge !

start-as-experiment.md
gruetter: Added section on related work.

PR #67, opened 2017-03-31
Update pattern nomenclature

nyeates:
New nomenclature and thoughts behind the maturities and states of patterns. This content was well-thought over before being brought here; it was created from a few back and forth emails and looking at all patterns states in the google doc


meta/pattern-states.md
```diff
@@ -1,13 +1,37 @@
...
-
-* types of patterns: donut, idea, draft
-* draft states: in progress, draft complete, ready for review, ready for 2nd review, etc
-* # of reviews completed: none, one, two, ...
+<img src="/assets/img/patterns-steps-and-maturities.png" height="350" style="float:right" align="right" alt="Maturity and Review Steps as seen on Github">
 
+## Pattern Maturity
+* A - Donut (Lacks Solution)
+* B - Unproven (Ideas)
+* C - Proven (Draft)
```
NewMexicoKid, line 9: Is a Proven pattern always a Draft?
Maybe: C - Proven (Pattern)

NewMexicoKid (commented):
Overall, this looks pretty solid. I'm wondering if the .md write-up should be more steady state rather than explanatory of how this proposal arose. E.g., change it from first person to third person.

meta/pattern-states.md
nyeates: The words in parenthesis () were meant to explain the OLD terminology. That isn't clear. I will change to your suggestion

PR #72, opened 2017-05-24
Fix a simple typo

jwsloan:


nyeates:
Thx I will merge! Do our descriptions make sense? They might need some clarification and we appreciate any feedback...

jwsloan:
I think the descriptions are great. I'm hoping to submit a new one soon. And I'm looking to apply the idea to both a "Why InnerSource" presentation and "Why IS Product X" template. It seems like any IS product idea would have a problem it is trying to solve, a context it exists in, forces acting within that context, and a desired resulting context.

PR #84, opened 2017-12-01
Contained InnerSource

gruetter:
Hi guys. Triggered by a suggestion of @NewMexicoKid in [this PR](https://github.com/paypal/InnerSourcePatterns/pull/15), I rewrote the pattern to better reflect the scenario I was thinking of.


contained-innersource.md
```diff
@@ -0,0 +1,70 @@
...
+  development resources (headcount).
+* _Full scale InnerSource_ includes substantial effort for soliciting and 
+   managing contributions as well as onboarding new contributors, which is 
+   hard to plan for and my have some effect on the ability to control the 
+   timeline for completion and deployment.
+
+## Solution
+
+The BUs involved in the development of the new product each dedicate
+development resources develop the new product, apply the InnerSource working
```
NewMexicoKid: development resources **to** develop the new product, apply the InnerSource working

contained-innersource.md
```diff
@@ -0,0 +1,70 @@
...
+model and adopt InnerSource tooling to facilitate collaborative development but
+limit the scope of InnerSource practices to the project. No effort will be
+spent on facilitating and managing contributions from outside of the project.
+The application of InnerSource is _contained_ to the project. The project may
+but is not required to share the developed artefacts outside of the project.
```
NewMexicoKid: Might be better to phrase it as: The project may share the developed artifacts outside the project but is not required to do so.

contained-innersource.md
```diff
@@ -0,0 +1,70 @@
...
+## Forces  
+
+* The product's importance to company revenue and the committed feature content
+  and dates require a development paradigm that provides known, stable
+  development resources (headcount).
+* _Full scale InnerSource_ includes substantial effort for soliciting and 
+   managing contributions as well as onboarding new contributors, which is 
+   hard to plan for and my have some effect on the ability to control the 
+   timeline for completion and deployment.
+
```
NewMexicoKid: * Code visibility/project transaparency is orthogonal to this particular Pattern. Leaving the code visible (the way other InnerSource projects would be) leaves the door open for future, more extensive collaboration opportunities, but the project may view this as possibly distracting their limited resources if unsolicited contributions are submitted. Keeping the code closed to participating members of the project doesn't activate the _Full scale InnerSource_ benefit of possibly higher quality due to having more eyes on the code but might satisfy potential project needs for secrecy (e.g., there may be legal or contractual obligations for the company to fulfill).

contained-innersource.md
```diff
@@ -0,0 +1,70 @@
...
+_Contained InnerSource_ effectively allows and facilitates the
+development collaboration of multiple BUs where existing traditional
+development infrastructure and processes would not have worked efficiently.
+
+Product development can be successfully completed. Even though _full scale
+InnerSource_ was not implemented and the full benefits from InnerSource
+practices, such as exemplary documentation which facilitates reuse beyond the
+project context, are not realized, the projects success still lends credibility
+to the concept of InnerSource.
```
NewMexicoKid: I think the principal benefit would be maybe better expressed as:
... the project's success still achieves several benefits that wouldn't have been possible through traditional development. In addition, it lends credibility to the concept of InnerSource.

contained-innersource.md
```diff
@@ -0,0 +1,70 @@
+## Title  
+
+Contained InnerSource
+
+## Patlet
+
+Apply InnerSource methods to facilitate collaboration in a cross divisional 
+project but don't invest in soliciting contributions from outside of that
+project.
```
jamestharpe: How is this different from traditional walled-off source?

contained-innersource.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Problem  
+
+Traditional development practices do not work as effectively for cross
+divisional projects as do InnerSource style practices. However, Management
+does not support adopting InnerSource practices because it regards efforts for
+soliciting and facilitating contributions from outside of the project as
+detrimental for the required efficiency.
```
jamestharpe: If management doesn't support it, shouldn't we build tools to help convince them rather than sacrifice the benefits of having internally open repositories? Will a hobbled version of InnerSource convince management of anything?

contained-innersource.md
jamestharpe: > the project may view this as possibly distracting their limited resources if unsolicited contributions are submitted.

Perhaps the best solutoin then is to propose a small experiment. For instance, leave the repo visible but don't be overly concerned with soliciting contributions. If it becomes a distraction, it can easily be locked down, if contributions turn out to be useful then the team can start to implement small steps to encourage further contribution.

contained-innersource.md
rrrutledge: Oftentimes traditional walled-off source allows anyone within the walls to talk to each other, but there's still just one team that adds code to a given project.  All others must get their feature ask or bug fix request prioritized into that team's backlog to get what they want.  The approach suggested here doesn't involved any more teams, but uplevels all teams involved to commit pull requests to a project instead of just feature requests.

contained-innersource.md
rrrutledge: I think it's fair to have more than one possible solution that addresses the same problem.  @NewMexicoKid would this just be a separate pattern with the same problem statement?

rrrutledge (commented):
Thanks for these comments, @jamestharpe!

contained-innersource.md
jamestharpe: Isn't this more of a "step toward" InnerSource, rather than a "pattern of" InnerSource? Do we make such distinctions? I worry that if a pattern like this is adopted, it can be seen as "good enough" by supporters of the status-quo and work against patterns that actually resemble open source.

rrrutledge: > Isn't this more of a "step toward" InnerSource, rather than a "pattern of" InnerSource?

I think so.  To be fair, I think that the *Resulting Context* calls that out clearly.

> Do we make such distinctions?

We haven't learned how to be that formal about it, yet.

> it can be seen as "good enough"

That's true and a good point.  @NewMexicoKid is there a place in the pattern format to put caveats or known "gotchas" with a particular pattern?

contained-innersource.md
spier: Fixed.

contained-innersource.md
spier: Adopted this suggestion.

contained-innersource.md
spier: Adopted.

contained-innersource.md
spier: In my interpretation of this pattern **Contained InnerSource** can indeed be seen as a "step towards" InnerSource. However I don't see this as a problem, as a step is better than no step. And if the situation really doesn't allow for more, I would take it :)

One other angle to look at this is to consider the  **Contained InnerSource** idea like a pilot project where the concept can be tried out. One hope is that the developers that experience it may advocate for it more strongly afterwards, and if they do they can use the specific experiences that they have made as examples that the decision makers can more easily relate to.

spier:
I have updated this pattern to the latest version of our pattern template and worked in some of the feedback from this PR.

As the pattern doesn't have a very specific **Known Instances** yet, I considered this pattern to be Level 1 (initial), rather than Level 2 (Structured). @gruetter Shall we keep this as Level 1 for now, or can you say more about the Known Instance?

Also @jamestharpe and @rrrutledge if you have seen this pattern in action since you commented on this in 2019, or want to leave any further ideas on this PR, please do.

Personally I would suggest to merge this pattern to the main line as is, so that we can get more eyeballs on in from the rest of the InnerSource Commons community.

contained-innersource.md
NewMexicoKid: > That's true and a good point. @NewMexicoKid is there a place in the pattern format to put caveats or known "gotchas" with a particular pattern?

Those go into the Resulting Context section; and it is not uncommon for significant "gotchas" or issues identified there to point to further patterns that could then be applied.

contained-innersource.md
NewMexicoKid: Yes. There can definitely be different solutions that address the same problem or that may apply when different contexts or forces are at play.

NewMexicoKid (approved):
The cleaned up pattern looks okay to me. Thanks for the effort, @spier , and thanks to @jamestharpe and @rrrutledge for the questions and comments.

spier:
Thanks for the feedback. Let's give this another week or so to solicit further feedback, and merge the best possible version.

patterns/1-initial/contained-innersource.md
```diff
@@ -0,0 +1,49 @@
+## Title  
+
+Contained InnerSource
+
+## Patlet
+
+Apply InnerSource methods to facilitate collaboration in a cross divisional project but don't invest in soliciting contributions from outside of that project.
```
spier, line 7:
```suggestion
Apply InnerSource methods to facilitate collaboration in a cross-divisional project but don't invest in soliciting contributions from outside of that project.
```

README.md
```diff
@@ -73,6 +73,7 @@ The below lists all known patterns. They are grouped into three [maturity levels
 * [Assisted Compliance](patterns/1-initial/assisted_compliance.md) - *Helping repo owners be compliant by writing their CONTRIBUTING.md for them as a pull request.*
 * [Open Source Trumps InnerSource](patterns/1-initial/open-source-trumps-innersource.md) - *Developers disregard InnerSource projects because they consider open source projects to be superior. Introducing a required evaluation of InnerSource projects before choosing an open source project increases the likelihood of the InnerSource projects to be adopted.*
 * [Transparent Governance Levels](patterns/1-initial/governance-levels.md) - *There are projects in multiple stages of InnerSource adoption. Contributors get confused when working with projects that are at different stages.*
+* [Contained InnerSource](patterns/1-initial/contained-innersource.md) - *Apply InnerSource methods to facilitate collaboration in a cross divisional project but don't invest in soliciting contributions from outside of that project.*
```
```suggestion
* [Contained InnerSource](patterns/1-initial/contained-innersource.md) - *Apply InnerSource methods to facilitate collaboration in a cross-divisional project but don't invest in soliciting contributions from outside of that project.*
```

PR #87, opened 2018-02-02
Cross-team Project Valuation

rrrutledge:
It's hard to sell the value of cross-team, inner sourced projects that don't provide a direct impact on company revenue.  Here's a pattern for constructing a data-driven valuation model for a cross-team project in any domain.  It will help you to represent your project in a data-driven way that both articulates its value and amplifies it.


gruetter:
FYI @rrrutledge : I've added a placeholder pattern on master named `crossteam-project-valuation.md`, so that we can keep links to this pattern intact once we merge (as per our decision in the patterns meeting from Feb 16th).

crossteam-project-valuation.md
```diff
@@ -0,0 +1,101 @@
...
+# Context
+
+* You are responsible for a cross-team that serves as a platform for others in the company.
+* The cross-team project does not deliver any direct value for company revenue.
+
+# Problem
+
+Cross-team projects can potentially have a very large impact on the company yet are notoriosly-difficult to represent in a data-driven fashion.
+As a result, it is easy and common to either pursue projects that does not provide real value or to underfund what would otherwise produce great value.
+Setting in place a pattern and model of how to value such projects gives us the tool that we need to focus and amplify high-value collaboration for the company.
```
gruetter: Should probably be reformulated as a problem  (absence of such a model) or moved to the solution section.

crossteam-project-valuation.md
```diff
@@ -0,0 +1,101 @@
...
+Here's a data-driven way to represent your project that both articulates its value and amplifies it.
+
+# Context
+
+* You are responsible for a cross-team that serves as a platform for others in the company.
+* The cross-team project does not deliver any direct value for company revenue.
+
+# Problem
+
+Cross-team projects can potentially have a very large impact on the company yet are notoriosly-difficult to represent in a data-driven fashion.
```
gruetter: It would be helpful to explain why that is so notoriously difficult, @rrrutledge. It's not obvious to me ;)

crossteam-project-valuation.md
```diff
@@ -0,0 +1,101 @@
...
+As a result, it is easy and common to either pursue projects that does not provide real value or to underfund what would otherwise produce great value.
+Setting in place a pattern and model of how to value such projects gives us the tool that we need to focus and amplify high-value collaboration for the company.
+
+# Forces
+
+Projects need to show value (objective or subjective) to company leadership in order to be funded.
+
```
gruetter: I would have expected more forces here, which explain the reason for the problem being there. Some ideas:
- Explain what makes valuation so difficult (cross-team nature?)
- Explain how a company fails to support the team absent valuation?!

crossteam-project-valuation.md
```diff
@@ -0,0 +1,101 @@
...
+## Explanation
+
+Assemble a small team of subject matter experts in your domain.
+Using that team of experts, estimate 4 things about each consumer of your project output:
+  * How long does it take them to consume your project output?
+  * How long would it otherwise take them to home-roll the value of your project output for themselves?
+  * What percentage of your project output is actually useful for them?
+  * How much time on an ongoing basis (ideally per-use) would they otherwise spend maintaining their home-rolled solution?
+
+When making these estimations, it's impossible to know with high precision _exactly_ how long any activities take.  That's not your goal.
```
gruetter: Nitpicky comment: I believe you are talking about _accuracy_, rather than _precision_ here.

crossteam-project-valuation.md
```diff
@@ -0,0 +1,101 @@
...
+Rather than exactness, you should strive to _**set a worst-case bound**_ on these estimates.
+The idea is for the group of experts to be able to say to each other, "We don't know exactly how long it would take, but we can all agree it's _at least_ this much."
+Specifically, you should estimate a *maximum* reasonable time to consume your project output and *minimum* reasonable times for consumers to otherwise home-roll, use and maintain their own solutions.
+
+> One note about cost-to home-roll.  The cost to home-roll a solution is NOT necessarily (very unlikely, in fact) the same as the cost of making a shared solution.
+Oftentimes for the same functionality the modularity and quality involved in buliding a cross-team, shared solution makes it a noticeably higher investment than a quick, hardcoded implementation used just once.
```
gruetter: Indeed. If I remember right from my software product line times, the rule of thumb was to not make something reusable unless it was used at least three times.

crossteam-project-valuation.md
```diff
@@ -0,0 +1,101 @@
...
+
+1.  The raw hours saved by the formula above.  Since we're all sure that the formula will produce a number that is below the true number of hours saved, you can have confidence that funding the project up to that point is a sure win for you.
+1.  The amount of time that it takes to support inner sourced contributions to cross-team projects.  Since the contributor would likely have done the work anyway in a one-off fashion, it is worth it to fund the time it takes to faciliate their work going into a shared location.
+1.  Whatever feels good to you.  One intentional side-effect of having a valuation formula is that it  naturally forces measurement of the key points of usage that provide value to consumers.
+Those measurements can be understood and consumed in their raw form to provide you with a gut-feel idea of how valuable is the project.
+
+Some may be concerned about the lack of precision in this  valuation approach.  It's okay for this process to not give  an exact measurement.  It just needs to be accurate enough to accomplish 2 purposes:
```
gruetter: again, I propose to `s/precision/accuracy/`

gruetter (approved):
Excellent pattern, @rrrutledge ! I really like this approach. One idea that came to mind reading it is to actually use a method from risk management: simple triangular distributions (gathered in subject matter expert group estimations) and Monte Carlo analysis. That way, you do not end up with just one number but rather a distribution and you can adjust based on your confidence. I have used this for project planning in the past and it has worked like a charm. Also, it appears very _sciency_ and managers liked it ;).

rrrutledge:
> you do not end up with just one number but rather a distribution

Yes, I've talked about similar ideas of a range.  With this pattern we're trying to hit the lower bound of the range.  What I've found is that if you get a lower bound that is low enough it becomes absurd for anyone to try and argue with it.  At the same time, even that absurdly low valuation level it usually turns out to show that we should be investing more in this kinds of cross-team projects because they are _so_ valuable.  You end up with what you want - an irrefutable, data-driven approach that justifies more investment.

crossteam-project-valuation.md
rrrutledge: > Explain how a company fails to support the team absent valuation

I'm not sure this item needs more explanation.  Isn't it fair to assume as a rule that companies need to be able to ascribe value to an effort in order to fund it?

crossteam-project-valuation.md
rrrutledge: Removed from here and moved to to the solution.

crossteam-project-valuation.md
rrrutledge: Removed the strong word, _notorious_, and moved this discussion to **Forces**.

crossteam-project-valuation.md
rrrutledge: > Explain what makes valuation so difficult

Added some explanation.

crossteam-project-valuation.md
rrrutledge: You're right - good point!

crossteam-project-valuation.md
rrrutledge: Makes sense.

NewMexicoKid:
Is this ready for a IS Patterns meeting review? I noticed no labels have been set...

rrrutledge:
> Is this ready

I don't know.  We were talking about the topic in a previous _Patterns_ meeting, and I said that I would share the content.  I'm not sure if other folks want to try to take it over the line or not.  I'm fine either way.

spier:
@rrrutledge this pattern looks interesting to me. 

Questions:

- What is preventing this from go into `master` in one shape or form, even if it might still need more work?
- Which "Review Steps" is this at?

My suggestion would be to already merge it to master, even if it is not fully reviewed yet. In `master` it will be easier for others to read this pattern than in this PR.

rrrutledge:
> My suggestion would be to already merge it to master

Thanks for advocating for this work!  The process around this working group has really broken down - maybe bring up your idea in the [#innersource-patterns](https://app.slack.com/client/T04PXKRM0/C2EFRTS6A) channel for general consensus?

crossteam-project-valuation.md
spier: @rrrutledge looks like this conversation can be resolved?

lenucksi: Looks resolvable unless we want to codify the rule of thumb knowledge here.

crossteam-project-valuation.md
lenucksi: Can this be resolved or are there open points left?

lenucksi (commented):
Looks good. States that is proven in multiple domains. Do we get 3 of them together? One is explicitly listed with Nike.

lenucksi:
Merging this in for now as it already resides in 2-structured. Please add additional pull requests against this file to further this pattern. @rrrutledge @gruetter.

rrrutledge:
Thanks, Johannes

Issue #89, opened 2018-02-19
pattern/service-vs-library


MaineC:
I would like to propose an idea for an inner source pattern - not sure if others have seen similar issues in their organisation.

Imagine the following situation: We are looking at an organisation that values devops as an approach to software development and deployment. Teams are responsible end to end: Starting with requirements engineering, software development down to deployment and maintenance, support etc.

Team A has developed service lovelyTree. Team B has a business issue, that could be solved by re-using lovelyTree, except they need a few modifications. When cross team collaboration and inner source is brought to the table, both teams are reluctant to collaborate - it's unclear who should get the wakeup call for the service, also error handling and communication chains are different for both teams.

The solution is to remind the team that this is about inner source, not inner deployment: Much like in open source, the same source code can be deployed in more than one context while still sharing the bulk of the code in one common service (or one common library that encapsulates the business logic).

This is related to the 30-day-warranty pattern, except it gives a bit more flexibility in a micro-services world.

Risk: This pattern breaks in a setup where one team has a vested interest in providing their software as a common service to others, e.g. when trying to collect data to feed into some machine learning training pipeline.

rrrutledge:
Sounds interesting

NewMexicoKid:
Isabel, I would encourage you to write this up in a pattern format.

chchrist:
@MaineC we have a team that developed a service that acts as a facade for another third party SaaS so in the future we could switch to a new same type SaaS without much overhead. Now other teams want to start using this service but the original team does not want to be responsible for problems that other teams might introduce. Does the above pattern fits this case? Does that mean that each new team will have to deploy their own service that does something similar?

lenucksi:
> @MaineC we have a team that developed a service that acts as a facade for another third party SaaS so in the future we could switch to a new same type SaaS without much overhead. Now other teams want to start using this service but the original team does not want to be responsible for problems that other teams might introduce. Does the above pattern fits this case? Does that mean that each new team will have to deploy their own service that does something similar?

Hey @chchrist, thanks for bringing this up! :+1: 
I think given the short form of "InnerSource not InnerDeployment" this might be interesting. There is also a mention of the "30 days of warranty" pattern in the linked PR discussion that might be of interest to you.

I've also posted a link to your question to our [Slack](https://innersourcecommons.org/slackinvite/) - you are very much invited to join there for more discussion.

PR #90, opened 2018-02-21
Add draft for service-vs-library pattern.

MaineC:
As requested in #89 this is a first draft of the service-vs-library pattern that I observed in DevOps teams.

Feedback, suggestions etc. welcome.

### DIFF
--- a/service-vs-library.md
+++ b/service-vs-library.md
@@ -5,8 +5,9 @@ Service vs. library: It's inner source, not inner deployment
 Teams in a DevOps environment may be reluctant to work across team boundaries on
 interdependent code bases due to ambiguity over who will be responsible for
 responding to service downtime. The solution is to realize that often it's
-possible to deploy the same service in independent environments with separate
-escalation chains in the event of service downtime.
+possible to either deploy the same service in independent environments with
+separate escalation chains in the event of service downtime or factor a lot of
+shared code out into one library and collaborate on that.
 
 ## Problem
 
@@ -15,7 +16,7 @@ feature end-to-end: From the customer down to deployment, maintenance and
 support. This poses a challenge when working across team boundaries: Escalation
 chains may not be the same for errors happening in either team. Coupling
 source code and deployment leaves the teams with the question of who is
-responsible for on-call support in the event of errros. As a result teams are
+responsible for on-call support in the event of erros. As a result teams are
 reluctant to join forces even if there is significant overlap in requirements.
 
 ## Context
--- a/service-vs-library.md
+++ b/service-vs-library.md
@@ -1,0 +1,82 @@
+## Title
+Service vs. library: It's inner source, not inner deployment
+
+## Patlet
+Teams in a DevOps environment may be reluctant to work across team boundaries on
+interdependent code bases due to ambiguity over who will be responsible for
+responding to service downtime. The solution is to realize that often it's
+possible to deploy the same service in independent environments with separate
+escalation chains in the event of service downtime.
+
+## Problem
+
+When teams are working in a DevOps environment developers are responsible for a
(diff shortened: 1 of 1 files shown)


service-vs-library.md
```diff
@@ -0,0 +1,82 @@
...
+escalation chains in the event of service downtime.
+
+## Problem
+
+When teams are working in a DevOps environment developers are responsible for a
+feature end-to-end: From the customer down to deployment, maintenance and
+support. This poses a challenge when working across team boundaries: Escalation
+chains may not be the same for errors happening in either team. Coupling
+source code and deployment leaves the teams with the question of who is
+responsible for on-call support in the event of errros. As a result teams are
```
NewMexicoKid: typo: errros -> errors

rrrutledge:
This writeup is good, Isabel.

service-vs-library.md
```diff
@@ -0,0 +1,82 @@
...
+Decouple responsibility for source code from deployment: Both teams work to
+identify exactly where there is overlap and synergies.
+
+Only shared source code is kept as part of the InnerSource project with shared
+responsiblity.
+
+Decouple configuration and deployment pipelines from actual business logic.
+Establish a second deployment of the service for the second team.
+
+Treat the common base as a library that is used by both teams with shared code
```
rrrutledge, line 54: The **Patlet** talks about `deploying the same service in independent environments` yet the solution talks about factoring out shared service code into a library, which becomes the focus of collaboration (not the service).  I feel like these two sections of the document are incongruent with each other.  Are the teams collaborating on the source code for a library or a service?

MaineC:
@NewMexicoKid Thanks for pointing out the typo.

@rrrutledge You're right - these two are incongruent. I've changed the Patlet to be better in line with the proposed solution. Deploying the same service twice with zero modification would presume that deployments are standardised across teams. While this may be true in some cases, more likely than not independent teams will have different deployment environments. Thank you for pointing out that flaw.

service-vs-library.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Problem
+
+When teams are working in a DevOps environment developers are responsible for a
+feature end-to-end: From the customer down to deployment, maintenance and
+support. This poses a challenge when working across team boundaries: Escalation
+chains may not be the same for errors happening in either team. Coupling
+source code and deployment leaves the teams with the question of who is
+responsible for on-call support in the event of errors. As a result teams are
+reluctant to join forces even if there is significant overlap in requirements.
```
rrrutledge, line 20: Another way of getting around this problem is [30-day warranty](https://github.com/paypal/InnerSourcePatterns/blob/master/30-day-warranty.md).  Could be good to mention that somewhere in this document.

rrrutledge (commented):
Seems pretty clear to me.  Just one comment.

Also, I accidentally made a commit (`cbb2bb2`) directly to this branch that I meant to make as a pull request suggestion to the branch.  If you don't like it then feel free to revert.

MaineC:
@rrrutledge I didn't even know that committing to a branch suggested as PR would be possible for the maintainers of the original repository. Learnt something new there. Also, thanks for the corrections you included in your commits.

service-vs-library.md
rrrutledge: Thanks.

rrrutledge (approved):
Looks good to me.

MaineC:
@rrrutledge thanks for the approval. What does the remaining process for this pull request look like?

NewMexicoKid (approved):
I think this is solid enough to move into the accepted patterns in the main branch. We can pursue a second review at need if people reading the pattern open a PR with suggested changes.

PR #93, opened 2018-03-27
Add "Related Patterns"

rrrutledge:
I like Georg's inclusion of it in the [start as experiment](https://github.com/paypal/InnerSourcePatterns/blob/master/start-as-experiment.md) document and thought to include it here.


rrrutledge:
Thanks, Tim.

PR #96, opened 2018-04-02
Place each sentence on its own line

rrrutledge:
Makes review and commenting easier.


NewMexicoKid (approved):
Good suggestion, Russ.

PR #104, opened 2018-08-27
Added revised version of community reviewed innersource-portal pattern

stephenmccall:


innersource-portal.md
```diff
@@ -0,0 +1,54 @@
+## Title 
+
+* InnerSource Portal
```
NewMexicoKid: Very minor: if there is just text in a section (like a one-line title in Title), it's probably better not to put that text as a bullet.

rrrutledge (approved):
I like this writeup.  I'd still love to git more crisp as to how patterns in this community use `Context`, `Forces`, and `Resulting Context`, but I'm still up-leveling here myself in this area.

innersource-portal.md
```diff
@@ -26,11 +25,14 @@ You are attempting to establish an InnerSource practice within your organization
...
+#### Prerequisites
+
 * Managers have given tacit acceptance that their employees should take part in InnerSource projects
-* InnerSource project owners are seeking a way to increase participation in their projects
 * A shared source control management system is in use which provides programmatic access to the contents of the repositories it hosts
-* The source control management application in use has limited search capabilities
-* There is a department within the organization with the responsibility of promoting InnerSource collaboration 
+* There is a department within your organization with the responsibility of promoting InnerSource collaboration
+
+Your organization is interested in adopting an InnerSource work style and InnerSource project owners are seeking a way to attract audiences to their projects. However, they are limited by the communication channels available to them through which they could advertise to potential contributors.  Compounding this problem is the fact that the shared source control management application in use has such limited search capabilities that even developers in search of InnerSource projects find it frustrating to locate them.
```
rrrutledge: I think that this paragraph should go above **Prerequisites** otherwise it looks like this content is part of the prerequisites.  It's actually the part of the context that is going to change (right)?

rrrutledge (changes requested):
One question - and suggestion if I am understanding things right.

rrrutledge (approved):
Yes!!!!

NewMexicoKid (approved):
Looks good!

PR #127, opened 2020-03-14
[docs] adding more explanation for Pattern Maturity and Review Steps.

spier:
Adding more explanation for Pattern Maturity and Review Steps.
Might help new contributors, as well as readers of this repo, what these concepts mean.

It looks like the rewrite of this nomenclature happened in 03/2017. I thought it was fair to assume that these concepts are now the ones that are used. Therefore I tried to focus the wording a bit more on "these are the states now", rather than "these are the changes that were made".

Potentially the whole section "Explanation of Changes" can be removed but that was too disruptive of a change for me to make, so I kept it there. Looks like the section might hold some historic knowledge that is still useful.


rrrutledge (approved):
Fantastic!

NewMexicoKid (approved):
Looks good.

Issue #142, opened 2020-03-27
Establish Automated Checks for Repo Contents


maxcapraro:
While this is not a source code repository, there are still some classes of bugs that we can avoid. We could establish some automated checks. From the top of my head:
- MD linting
- Dead link checker
- Spell checker (via Russ)
- ... (potentially others)

spier:
Nice ideas. How does the output of "MD linter" look like? I have never seen one in action?

In addition a check for the require headilnes in pattern files would be nice. Basically a help for new authors to get the pattern format right. This is only really practicable once we have moved the patterns to a separate folder thought.

I think as a first step the 'dead link checker' would be most useful.
It's important that it can deal with relative links inside the markdown though (e.g. from README.md to a pattern file, as in the. case where we discovered this issue).

Two GitHub Actions that we could try:
* https://github.com/marketplace/actions/link-checker
* https://github.com/marketplace/actions/markdown-link-check

spier:
I tried [link-checker](https://github.com/marketplace/actions/link-checker).
It works ok. See sample output below.

If we were to start with this I think we should limit it to README.md initially, as otherwise we have way too many broken links already :)

```
Link checker output file: link-checker/out.md
README.md
	OK	/meta/meetings.md
	OK	30-day-warranty.md
	OK	CONTRIBUTING.md
	OK	bad-weather-for-liftoff.md
	OK	change-the-developers-mindset.md
	OK	common-requirements.md
	OK	contracted-contributor.md
	OK	crossteam-project-valuation.md
	OK	dedicated-community-leader.md
	OK	defeat-hierarchical-constraints.md
	OK	developer-incentive-alignment-for-innersource-contribution.md
	OK	gig-marketplace.md
	OK	http://bit.ly/innersource_patterns_videos
	OK	http://hillside.net/index.php/a-pattern-language-for-pattern-writing
	OK	http://innersourcecommons.org
        ...
	OK	start-as-experiment.md
	ERROR	http://creativecommons.org/licenses/by-sa/4.0/
		 (HTTP error 521)
	ERROR	i-am-not-here.md
		Stat i-am-not-here.md: no such file or directory
```

spier:
I tried a markdown linter on the `README.md` file, just too see how that would look like.

If we don't check for line length (which the linter complaints about all over the place), we would still have these errors:

```
$ markdownlint README.md
[...]

PR #145, opened 2020-04-04
Add link checker only for README.md

spier:
Adds one check as discussed in #142. 

This introduces a GitHub Action that checks all links in `README.md` on every push. Links are checked using https://github.com/raviqqe/liche.

This will check both local links (files), as well as http links.

If any link isn't reachable, or the filenames linked are wrong, the check will fail and block a PR to be merged into master (assuming that this is how the protection rules for master are configured).

## Implementation Details:

The link to `http://creativecommons.org/licenses/by-sa/4.0/` is ignored by this check, as somehow that returns 521 even though it loads correctly in the browser.

Error response was:

```
	ERROR	http://creativecommons.org/licenses/by-sa/4.0/
		 (HTTP error 521)
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,17 @@
+# from: https://github.com/marketplace/actions/link-checker
+
+name: Link Check on README.md
+
+on: push
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Link Checker
+      id: lc
+      uses: peter-evans/link-checker@v1
+      with:
+        args: -v -d . -x http://creativecommons.org/licenses README.md
+    - name: Fail if there were link errors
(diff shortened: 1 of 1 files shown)


spier:
@lenucksi you will like this one :)

lenucksi:
> @lenucksi you will like this one :)

Indeed I do @spier :smile:.
I'll merge it in for now, we can still modify its settings or triggers once we see the first results.

lenucksi:
Looks like https://isc-inviter.herokuapp.com/ also needs an exception. It is reported as `HTTP 503` although it works fine. 
@spier Would you mind adding a PR for that? I'm happy about matching PRs from everyone else, too. :smiley:

spier:
Strange, I did not get this issue. Will try again locally.

spier:
Here you go:
https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/146

PR #146, opened 2020-04-05
Add Link Checker exception for isc-inviter.herokuapp.com

spier:
I cannot test if this really works, because for whatever reason the link check for isc-inviter.herokuapp.com does not fail for me.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,6 @@ jobs:
       id: lc
       uses: peter-evans/link-checker@v1
       with:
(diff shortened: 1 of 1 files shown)


lenucksi (approved):
LGTM

lenucksi:
Run https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/71877761 just finished green / clean, I think we're good here then @spier.
Let's see if we get the desired effects on the next PRs or if we need to add PR as a run condition.

PR #148, opened 2020-04-12
Pattern Template: Marking Authors and Acknowledgements as optional

spier:
Marking Authors and Acknowledgements as optional, because based on the description they are.

@lenucksi this might be a super quick one to merge as I don't think there is anything controversial in here.


lenucksi (approved):
That's indeed a quick one. Thanks & LGTM

Issue #149, opened 2020-04-13
Establish `.github/codeowners` file with the existing TC's handles.


lenucksi:
We've listed a few Trusted Committers in our `TRUSTED-COMMITERS.md` file. We've discussed the benefit of using the GitHub automation around this.
Hence, this issue is about enabling this.
- `.github/codeowners` file with the listed persons handles in it. 
- That would this would give [a bit of automation](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners).

_Originally posted by @lenucksi in https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/141_

spier:
One thing that recently learned:

If all users that you want to enter into the `CODEOWNERS` file are part of the org that owns this repo, then you can also put them all in a team and then put that team name into CODEOWNERS.

That way you can route PR reviews automatically to individual members of that team, rather than to alert everybody. See:
https://help.github.com/en/github/setting-up-and-managing-organizations-and-teams/managing-code-review-assignment-for-your-team

However that only works if all user in question are part of this org. If you also want to add outside collaborators to the file, then that wouldn't work 100%, as outside collaborators cannot be part of teams in GitHub.

spier:
Perfect. Done! 
(Just assuming that this will work, as I have tested this).

lenucksi:
Done by means of #151.

PR #151, opened 2020-04-24
Adding CODEOWNERS file. Also linking adding clickable links to the trusted commiters.

spier:
Adding CODEOWNERS file. Also linking adding clickable links to the trusted commiters.

Implements #149


spier:
Here you go @lenucksi :)

lenucksi (approved):
Thanks for creating this! Looks good to me.

PR #153, opened 2020-04-25
Move meta info from main README to /meta

spier:
This PR simplifies the main `README.md`, by moving the block that is about maintaining the patterns to a `meta/README.md`.

Also moves `pattern-system.md` to `meta/`.


spier:
Alright, so the CODEOWNERS definitely work, given all the automatic review assignments to this PR :)

However the link check is not running automatically. It might only be triggered by a push to master right now actually? Not sure, as I don't know GitHub Actions in that much detail.

NewMexicoKid (approved):
Looks okay. Not 100% convinced that we need to simplify the README (I think it may be more of a personal preference whether to have all the information in a single file or require those who are interested to do a further click), but I'm okay with the proposal.

spier:
Thanks for the feedback @NewMexicoKid and for the approval. I can certainly see that argument for or against multiple READMEs.

I could have explained the purpose of my change better. Let me try:
Right now our main README is the starting point of the ISC Patterns for all users (*readers* and *contributors*). You could say the main README is the homepage for ISC Patterns :)

Assuming that the majority of the users will be just *readers*, I would like to make the main README as simple and short for the *readers* as possible. The more complicated info is then available for the *contributors* in a separate place (`./meta`).

As this is already approved, do you or @lenucksi want to merge this?

maxcapraro:
> Right now our main README is the starting point of the ISC Patterns for all users (readers and contributors). You could say the main README is the homepage for ISC Patterns :)
>
> Assuming that the majority of the users will be just readers, I would like to make the main README as simple and short for the readers as possible.

This will likely change once we have releases of a patterns catalogue, e-book as PDF, interactive website, etc. But in the current state I fully agree with that observation!

maxcapraro (approved):
I agree with this observation of @spier:

> Right now our main README is the starting point of the ISC Patterns for all users (readers and contributors). You could say the main README is the homepage for ISC Patterns :)
>
> Assuming that the majority of the users will be just readers, I would like to make the main README as simple and short for the readers as possible.

This will likely change once we have releases of a patterns catalogue or e-book as PDF, interactive website, etc. But in the current state, I think, the changes proposed with this PR will improve the experience for readers.

spier:
Great. @lenucksi, ready to merge? :)

lenucksi:
> Great. @lenucksi, ready to merge? :)

Yes. Big green button time!.

> However the link check is not running automatically. It might only be triggered by a push to master right now actually? Not sure, as I don't know GitHub Actions in that much detail.

This is more or less as expected. And yes, your assumption is correct. We likely need to adapt the actions description to [trigger on PR](https://help.github.com/en/actions/reference/events-that-trigger-workflows)
Do you want to file a quick PR for this?
Side note: We can have as much GitHub Actions actions as we want, i.e. could split scripts to run on pushes to master from those running on PR. 

I also agree with the sentiment on separation for Contributors and Users brought forward by @spier . I also agree with @NewMexicoKid. 
And most cases I've answered "to split or not to split" based on the extent of the contributor focused material. We have quite a bit in that regard. Hence I'd go for splitting for now. And change things again when they need changing. :) 

On a different note (and l'd be very happy to deal with that in another PR, just wanted to mention it here too so that the context is created): Do we need and if so how do we integrate the works in #155 in these changes?

lenucksi (approved):
> Great. @lenucksi, ready to merge? :)

Yes. Big green button time!.

> However the link check is not running automatically. It might only be triggered by a push to master right now actually? Not sure, as I don't know GitHub Actions in that much detail.

This is more or less as expected. And yes, your assumption is correct. We likely need to adapt the actions description to [trigger on PR](https://help.github.com/en/actions/reference/events-that-trigger-workflows)
Do you want to file a quick PR for this?
Side note: We can have as much GitHub Actions actions as we want, i.e. could split scripts to run on pushes to master from those running on PR. 

I also agree with the sentiment on separation for Contributors and Users brought forward by @spier . I also agree with @NewMexicoKid. 
And most cases I've answered "to split or not to split" based on the extent of the contributor focused material. We have quite a bit in that regard. Hence I'd go for splitting for now. And change things again when they need changing. :) 

On a different note (and l'd be very happy to deal with that in another PR, just wanted to mention it here too so that the context is created): Do we need and if so how do we integrate the works in #155 in these changes?

maxcapraro:
> On a different note (and l'd be very happy to deal with that in another PR, just wanted to mention it here too so that the context is created): Do we need and if so how do we integrate the works in #155 in these changes?

I don't think there is much conflict. #153 makes sense - Thanks for contributing and merging this :)

PR #158, opened 2020-04-29
Fixes link to /meta folder

spier:
Fixes the broken link that I introduced in https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/153.

So I guess that's why we need the automatic linkchecker :)


maxcapraro (approved):
This change is good. I suggest to merge it.

lenucksi:
#instamerge

PR #160, opened 2020-05-05
Use h2 headlines (##) according to the Pattern Template

spier:
While working on a linter/validator for our pattern syntax, I found these two patterns that incorrectly use h1 rather than h2 headings.

This PR fixes that.


lenucksi (approved):
Great, thanks!

PR #161, opened 2020-05-06
Fixes link to pattern discover-your-innersource.md

spier:
The PR for this pattern had been merged some time ago, so it makes more sense to link to the markdown, than to the PR. Also the pattern has been renamed in the meantime.


lenucksi (approved):
Cool, thanks for the update.

PR #164, opened 2020-05-07
Remove links from patterns to wiki pages that don't exist any more

spier:
When we ported wiki pages into markdown files, we created links from the `.md` pattern files to the respective wiki pages. 

Now that those [wiki pages have been deleted](https://github.com/InnerSourceCommons/innersourcecommons.org/issues/134), these links are no longer working, and not needed anymore.

This PR removes these broken links from the `.md` pattern files.


lenucksi (approved):
Thanks, you are indeed right. Merging this.

PR #167, opened 2020-05-08
Rename 'Inner Source' to 'InnerSource' in README

spier:
I noticed that the spelling of `Inner Source` to `InnerSource`. 

There is probably not correct spelling per se, but at least within this repo we can try to be as consistent as possible. (side note: [wikipedia](https://en.wikipedia.org/wiki/InnerSource) has it as `InnerSource`.)

I checked which spelling we have been using in this repo, and `InnerSource` is the clear winner :)

```
$ ack -h -c "Inner Source" *.md
8

$ ack -h -c "InnerSource" *.md
293
```

Hence changed the occurrences of `Inner Source` to `InnerSource` in the main README, to make this entry point to our repo more consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,16 +74,16 @@ possible to either deploy the same service in independent environments with sepa
 * [Get Contributions Despite Silo Thinking](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/38)
 
 
-# What are Inner Source Patterns?
+# What are InnerSource Patterns?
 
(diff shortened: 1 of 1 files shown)


lenucksi:
> I noticed that the spelling of `Inner Source` to `InnerSource`.
> 
> There is probably not correct spelling per se, but at least within this repo we can try to be as consistent as possible. (side note: [wikipedia](https://en.wikipedia.org/wiki/InnerSource) has it as `InnerSource`.)
> 
> I checked which spelling we have been using in this repo, and `InnerSource` is the clear winner :)

Yes. It is.

> 
> ```
> $ ack -h -c "Inner Source" *.md
> 8
> 
> $ ack -h -c "InnerSource" *.md
> 293
> ```

Ah, I see a man of culture! :D I tend to use ["the silver searcher"](https://github.com/ggreer/the_silver_searcher) these days though, pretty much `ack`, just faster. 

> 
> Hence changed the occurrences of `Inner Source` to `InnerSource` in the main README, to make this entry point to our repo more consistent.

Thanks. Merging.

PR #177, opened 2020-05-19
Change headlines in README

spier:
I noticed that the main `README.md` does not have a title. This PR fixes this, and other issues related to headlines.

- add title to README
- reduce level of other headlines accordingly (all go one level down)
- change to one-level decrease per step
- remove some superfluous newlines


lenucksi (approved):
Nice, thanks! :D

Issue #192, opened 2020-08-11
Move Pattern "Reluctance to Receive Contributions" to markdown


spier:
The pattern [Reluctance to Receive Contributions](https://docs.google.com/document/d/13QDN-BpE_BixRFVGjao32n4Ctim0ROXAHbBWMBOijb4/edit) currently lives in a Google Doc. We would like to move it to a markdown file in this repo, so that it is easier for everybody to access this pattern.

@user-group it looks like you might know more about this pattern? 

Do you think you could convert this pattern to markdown?

**As an aside:**

We have reworked the [maturity levels][levels] of the InnerSource patterns.  Could you take a look which level this pattern might fall into? We currently have it under **Maturity Level 2: Structured** but I wasn't sure if that is correct.

Thank you for your help! 👍 

[levels]: https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/meta/contributor-handbook.md#maturity-levels
[format]: https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/meta/pattern-template.md

spier:
This work has been done. Closing the issue.

PR #199, opened 2020-08-15
New Pattern: reluctance to accept contributions

spier:
First conversion of pattern in gdoc into markdown.

Implements https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/192

If we manage to add all missing template pieces as part of this PR, then we can move the pattern to `patterns/2-structured/` as well.

### DIFF
--- a/patterns/1-initial/reluctance-to-accept-contributions.md
+++ b/patterns/1-initial/reluctance-to-accept-contributions.md
@@ -8,7 +8,7 @@ The core team that owns a shared asset is reluctant to adopt contributions of co
 
 ## Context
 
-An organizational unit (e.g. a business division or development team) is using an inner source product (a ‘shared asset’), and finds it is missing a certain feature that they need and would like to integrate into the shared asset. The core team that owns the shared asset is responsible for its design and maintenance. The potentially contributing organizational unit has no champion on the core team. 
+An organizational unit (e.g. a business division or development team) is using an InnerSource product (a ‘shared asset’), and finds it is missing a certain feature that they need and would like to integrate into the shared asset. The core team that owns the shared asset is responsible for its design and maintenance. The potentially contributing organizational unit has no champion on the core team. 
 
 ## Forces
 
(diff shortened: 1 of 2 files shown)


spier:
@user-group in the `## See Also` section this pattern here references other patterns, but I could not clearly identify which patterns those are.

- Support Warranty => probably [30 Day Warranty](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/patterns/2-structured/30-day-warranty.md)
- Negotiate Contributions => ???
- Architecture for Participation => ???
- Feature Advocate => ???

Could you help me identify if those patterns are "linkable" somewhere?
Otherwise we can remove them from this pattern maybe?

spier:
I believe that @user-group won't have time for this for the foreseeable future. 

Therefore I would now remove the links to the patterns that are unknown/undiscoverable, so that we can get this pattern shared with a wider audience here in this repo (as opposed to just having it as a gDoc).

Does that work for you @lenucksi ?

patterns/1-initial/reluctance-to-accept-contributions.md
```diff
@@ -0,0 +1,74 @@
...
+
+## See Also
+
+* Support Warranty - most likely [30 Day Warranty](../2-structured/30-day-warranty.md)
+* Negotiate Contributions (unknown pattern)
+* Architecture for Participation (unknown pattern)
+* Feature Advocate (unknown pattern)
+
+*Note:* 
+The last 3 patterns above were mentioned to be "under construction" at the time where this pattern was written. Right now we don't know where these patterns are.
```
spier, line 50: @lenucksi I decided to solve the problem with the unknown patterns like this now.

- linked to the pattern that we know
- left the other 3 in there without a link. maybe we can still resolve them when somebody invests time to level this up to Structured

spier:
In a previous version of the pattern (in the original gDoc) I found other pattern names listed under **##See Also**:
Not invented here, (TBW) Thirty-Day Support, (TBW) Trusted Committer, (TBW) Fear of Contributions

Maybe that info could help us resolve which patterns the authors had in mind when they wrote this?

patterns/1-initial/reluctance-to-accept-contributions.md
lenucksi: Sounds good to me. I've converted the other comment to an issue such that we have it to follow up for whoever might want to do that.

lenucksi:
Linkbot is unhappy about the open-org link here, which is still broken although the file is not touched in this PR leading to an unwarranted red check.

lenucksi (approved):
Thanks for the transport of the pattern. It's fine for 1-initial for now. Added #226 to follow-up on unresolved patterns references.

PR #201, opened 2020-08-16
update current mindmap to reflect patterns' new location

fwan2000:
Update mind map files to use the newly restructured pattern locations.


lenucksi (approved):
Great, thank you very much @fwan2000!  :smiley:

PR #203, opened 2020-09-02
Fixes for the Issue Tracker pattern

spier:
Some fixes for the Issue Tracker pattern. 

- fixes headline level
- changes order of sections (first the Problem, then the Context)
- adapts status to new maturity levels

@MaineC this does not contain any content changes, so hopefully a quick one to review.


lenucksi (approved):
Thanks. LGTM

PR #205, opened 2020-09-05
Calling out the parent folder '/patterns' explicitly

spier:
I noticed that we were just mentioning the sub-folders for different maturity levels, but forgot to mention that they are all placed inside of `patterns`.

This PR fixes the docs accordingly.


spier:
@lenucksi this is likely a no-brainer merge for you 👍

lenucksi (approved):
Indeed a no-brainer.

PR #208, opened 2020-09-15
Added SAP as a known instance

Michadelic:


lenucksi (approved):
LGTM, thanks. Validated at ISC Fall Summit 2020

PR #220, opened 2020-09-22
Standardize title section

spier:
While working on the automation for the book creation, I needed to parse out the pattern titles from the markdown files. 
I realized that some titles are using a different formatting, which makes the parsing harder and the book also looks a tiny bit inconsistent then.

In this PR I change this:
- remove italics from titles
- move aliases for pattern titles to a separate `Alias` section. This gives the pattern one clear **main name**. the aliases could later be used as separate links to the same pattern if we like to do that.

@lenucksi we don't have the "Alias" section header defined in the Pattern template. However I assumed that extra headers are fine to add, as long as they don't contract the other headers. 

Should I add a definition of the "Alias" section to the Pattern template file as well?

### DIFF
--- a/meta/pattern-template.md
+++ b/meta/pattern-template.md
@@ -33,10 +33,10 @@ Where has this been seen before? Helps to reinforce that this is a REAL pattern 
 
 May mention:
 * A particular business
-* Anonymized instances ex: "3 companies have proven that this is a good solution" or "A large financial services org...". 
+* Anonymized instances ex: "3 companies have proven that this is a good solution" or "A large financial services org...".
 
 ## Status (optional until merging)
-General pattern status is stored in GitHub's Label tagging - see any pull request. Note that this GitHub label tagging becomes less visible once the pattern is finalized and merged, so having some information in this field is helpful. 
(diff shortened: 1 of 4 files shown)


lenucksi:
> While working on the automation for the book creation, I needed to parse out the pattern titles from the markdown files.
> I realized that some titles are using a different formatting, which makes the parsing harder and the book also looks a tiny bit inconsistent then.
> 
> In this PR I change this:
> 
> * remove italics from titles
>
> * move aliases for pattern titles to a separate section (this gives the pattern one clear **main name**. the aliases could later be used as separate links to the same pattern if we like to do that)

Thanks, looks good to me.
> 
> 
> @lenucksi we don't have the "Alias" section header defined in the Pattern template. However I assumed that extra headers are fine to add, as long as they don't contract the other headers.
> 
> Should I add a definition of the "Alias" section to the Pattern template file as well?

If we assume this to be used consistently it should go into what currently is used for sort-of "documentary" purposes, i.e. the template. 
Also having this in a separate section will allow us to use this to index or other parser uses later on.

lenucksi (changes requested):
Should document the use of alias in the pattern template as illustrated above. Good otherwise.

spier:
Done!

lenucksi (approved):
LGTM

PR #168, opened 2020-05-09
Demo of Linter warnings

spier:
Draft PR for #159 to check how the annotations for the markdown linter would work.

PR #222, opened 2020-09-23
chore: Add Markdownlint to CI

nschonni:
- Turn off failing rules
- Add problem matcher to show errors in CI results
- Limit CI run to Markdown changes

### DIFF
--- a/.github/workflows/markdownlint.yml
+++ b/.github/workflows/markdownlint.yml
@@ -1,0 +1,32 @@
+name: Markdownlint
+
+on:
+  push:
+    paths:
+      - "**/*.md"
+      - ".markdownlint.json"
+      - ".github/workflows/markdownlint.yml"
+      - ".github/workflows/markdownlint-problem-matcher.json"
+  pull_request:
+    paths:
+      - "**/*.md"
+      - ".markdownlint.json"
+      - ".github/workflows/markdownlint.yml"
+      - ".github/workflows/markdownlint-problem-matcher.json"
+
+jobs:
+  lint:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.x
+    - name: Run Markdownlint
+      run: |
+        echo "::add-matcher::.github/workflows/markdownlint-problem-matcher.json"
+        npm i -g markdownlint-cli
+        markdownlint "**/*.md"
--- a/.github/workflows/markdownlint-problem-matcher.json
+++ b/.github/workflows/markdownlint-problem-matcher.json
@@ -1,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "markdownlint",
+            "pattern": [
+                {
+                    "regexp": "^([^:]*):(\\d+):?(\\d+)?\\s([\\w-\\/]*)\\s(.*)$",
+                    "file": 1,
+                    "line": 2,
+                    "column": 3,
+                    "code": 4,
+                    "message": 5
+                }
+            ]
+        }
+    ]
+}
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,22 @@
+{
+    "default": true,
+    "MD004": false,
+    "MD007": false,
+    "MD009": false,
+    "MD010": false,
+    "MD012": false,
+    "MD013": false,
+    "MD014": false,
+    "MD022": false,
+    "MD024": false,
+    "MD025": false,
+    "MD026": false,
+    "MD029": false,
+    "MD030": false,
+    "MD032": false,
+    "MD033": false,
+    "MD034": false,
+    "MD036": false,
+    "MD040": false,
+    "MD041": false
+}


spier:
@nschonni nice idea! Thanks for the contribution.

Where can I see how the results of this GH Action run would look like?

Also out of curiosity, is there an advantage to adding the markdownlint to the `main.yml` workflow, or could this as well go to a separate workflow file?

nschonni:
@spier I can turn on one of the failing rules if you want to see what it will look like

It could definitely go in a separate file if you want. Think I have one from another project that I could just copy over instead of editing the existing job if you want

spier:
@nschonni Cool. Yeah, let's do a separate workflow file then if you don't mind.
That will also help us to debug the GitHub Actions stuff (because at least I don't know those that well yet).

If you have a screenshot of how the failing markdownlint check would look like, that would also do.

Also will that check make the PR fail, or will it just issue a warning?

Thanks again for your help, I am sure this will be really good!

Btw in a separate PR (#168) I have been experimenting with a linter to check the patterns against our pattern template. Mostly just checking that the required headers exist and are in the right order. There I have also been experimenting with writing a custom rule for markdownlint. So I might pick you brain on that in the future if you don't mind :)

nschonni:
@spier I split it out and added a failing rule to see on the files tab at the bottom what it would look like when there are issues on the PR https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/222/files. Note that it will flag issues in files outside the PR as well as the ones in the PR, but the style you can see there is what it would show directly inline in the changed files.

I've flipped this to draft since there is the purposely fialing rule. If you're OK with this, I'll rebase off the commit and turn off the Draft bit. 

> Also will that check make the PR fail, or will it just issue a warning?

It would show the red X, but it wouldn't prevent merging unless you make the check required https://docs.github.com/en/github/administering-a-repository/about-required-status-checks

spier:
Nice, I like the checks!

When reading the issues that markdownlint reported inline, I was wondering how somebody that pushes a PR can figure out what the error means? e.g. `Hard tabs [Column: 1]`. Is it possible to point PR authors to the [official rules documentation](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md010---hard-tabs), or how to best educate them?

Lastly, do you think if it makes a difference if we check all markdown files in the repo vs just checking what is in `/patterns/*`? Currently this PR is checking all files in the rep, right?

spier:
Besides the specifics of this PR:
Do you happen to be in the InnerSource Commons Slack already? You can [join here](http://innersourcecommons.org/slack). In the `#innersource-patterns` channel there we are discussing all things related to InnerSource Patterns. So if you like to chat with likeminded folks, might be interesting for you :) 

In either case, I really like that markdownlinter thing already. Thanks so much for sending a PR 👍

nschonni:
> When reading the issues that markdownlint reported inline, I was wondering how somebody that pushes a PR can figure out what the error means? e.g. `Hard tabs [Column: 1]`. Is it possible to point PR authors to the [official rules documentation](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md010---hard-tabs), or how to best educate them?

I'm not seeing an option directly from this UI, but if they're using something like VS Code and the plug-in, I know it will offer a link to the docs. Do you want me to add a .vscode folder with the suggested extension? I know that won't cover everyone, but might be useful for those that do.

> Lastly, do you think if it makes a difference if we check all markdown files in the repo vs just checking what is in `/patterns/*`? Currently this PR is checking all files in the rep, right?

I would probably say to run it across everything since it applies to all Markdown files, and people will/do see the issues flagged in their IDE if they have a plug-in already installed

I'm going to rebase off the test commit, and I'll just add a snip of the example errors
image

spier (approved):
LGTM

spier:
I would opt for not doing anything IDE-specific yet. Let's just merge this as is and make our first experiences with the linter "in Production". 

Once we are more comfortable with it we can revisit if we want to make the checks any tighter (i.e. ignore fewer rules) and if we want to do anything else to make the linter more helpful for the majority of the contributors.

@lenucksi I think this one is good to be merged!

lenucksi (approved):
Thanks for creating this @nschonni! 

Regarding the execution I think running it against all Markdown files for now on _push_ and the files touched by the _pull request_ on pull request should be fine.

Regarding specific editors or IDEs I think one should - if at all - strive for something portable such as https://editorconfig.org/ Or just add a link to the documentation somewhere - might even be an `echo` statement on with the link to the docs on a failed check in the action.

Will merge this mid next week latest if I don't hear anything new.

spier:
@nschonni I created a custom linter with a slightly different purpose: 
To check if the patterns comply with the syntax defined in our pattern template. See https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/257.

I noticed that in my approach I am using an existing GitHub Action `avto-dev/markdown-lint@v1`, while in your approach you seem to be installing node and markdownlint-cli from scratch. I don't know that much about GitHub Actions and markdownlint, so I am just wondering if there is any advantage of one approach over the other?

nschonni:
I've never come across that project before, but it sounds like it's a Docker image ontop of the CLI with some sort of other layer on top.
Quick looking at the PR, and I think some of your custom rules are probably things that can be done using the built-in rules rather than writing your own

spier:
I tried to use the built-in rules but could not quite get it to work. But I would be thankful for any pointers that you might have, and happy to try it again.

What has proven difficult for our case that I want to check for mandatory headlines, and at least the built-in rule that I tried did not quite work.

nschonni:
https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md043

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,32 @@
+name: Markdownlint
+
+on:
+  push:
+    paths:
+      - "**/*.md"
+      - ".markdownlint.json"
+      - ".github/workflows/markdownlint.yml"
+      - ".github/workflows/markdownlint-problem-matcher.json"
```
spier, line 9: Are you sure that these paths are required?

spier:
Thanks for the pointer!
I did try out rule MD043 before but I couldn't quite get it to work the way that I wanted.

What I do remember was that:
1) Some patterns have minor differences in the spellings of some section headings (e.g. Author/Authors etc) and we have let those go so far. MD043 does not allow for multiple spellings of the same heading. => one solution would be to be more strict and fix the "offending" section headings
2) While we do suggest a certain template structure, we have been pretty flexible if people wanted to add additional headlines as well. I tried implementing this with the `*` syntax of MD043 but got stuck somewhere.

Might revisit this at some point, for now it is good if we have any form of syntax linter, as at least it will remind us not to forget about it :)

spier:
But let's change the topic back to your own linter and this PR actually:
It looks like it is ready to get merged as is, right?

I left one [comment](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/222/files#r549176443) inline about the way that the Action is configured.

It am mostly exploring where we can best place the markdownlint config files etc so that our two linters don't get into each others way.

.github/workflows/markdownlint.yml
nschonni: yeah, it makes sure the CI is triggered when any of the files in the pipeline are touched. Alternately, there could be no filter and it could run on every PR

spier: Got it.

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,32 @@
+name: Markdownlint
+
+on:
+  push:
+    paths:
```
spier, line 5:
```suggestion
on:
  push:
    branches:
      - "master"
    paths:
```

I am suggesting this change because without this limitation for triggering the workflow only on push to `master` Ii have seen duplicate GHA workflow executions before.

With this change the configuration would be similar to this other GHA that we use:
https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/.github/workflows/main.yml

spier (changes requested):
I am suggesting one change should prevent duplicate workflow executions. Other than that this looks great to me and I am happy to merge it.

.github/workflows/markdownlint.yml
nschonni: If you limit this to only pushes to the default branch, you won't get checks on PR branches ahead of time, which leads to churn after opening a PR. The filter can also be useful for jobs that really depend on paths in a particular branch.

spier: Ok. Then let's try it as is. 
I will resolve this thread and merge the PR.

spier (approved):
LGTM

spier:
Your linter implementation is now live in `master` 🎉 
Thanks for the great contribution, which will help us to produce more consistently formatted markdown in our patterns.

@nschonni just one clarifying question though:
When you [said](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/222#discussion_r549450890) "you won't get checks on PR branches ahead of time" did you mean that the Action won't get triggered before a PR is opened for the given branch?

nschonni:
> When you [said](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/222#discussion_r549450890) "you won't get checks on PR branches ahead of time" did you mean that the Action won't get triggered before a PR is opened for the given branch?

If you filter on the branch for the job, then yes, there would be no trigger. EX: your other job only triggers when things are merged into the default branch, or when an PR is opened, but not when you're pushing and working on `my-feature-branch`

spier:
Got it, thanks.

Issue #218, opened 2020-09-22
Basic spell checker for patterns


spier:
Patterns written straight in markdown format often have some minor spelling mistakes, as GitHub does not provide any automatic checks for that.

Therefore the patterns (and the book that we want to produce) would benefit from at least a basic spell check. That spell check could e.g. run when a PR is created. The spell checker should not make the CI run fail. I would rather use the results of the spell checker as recommendations to the author that they can decide to use, or not.

I found one spell checker at https://norvig.com/spell-correct.html but I am sure there are many.

lenucksi:
Added the [prosebot](https://github.com/apps/prosebot) to our repo. It should become active on PRs for .md and .txt now. 
There are three linters active I can envision that we might reduce them to the spellchecker only.

spier:
Nice nice nice. Looking forward to seeing the bot in action.

Feel free to close this issue already if you consider it solved (we can open it again if we want to do more at a later point).

lenucksi:
> 
> 
> Nice nice nice. Looking forward to seeing the bot in action.
> 
> Feel free to close this issue already if you consider it solved (we can open it again if we want to do more at a later point).

I'll close once I've seen that it works.

spier:
@lenucksi I just remembered prosebot, and that you added it. However in the last PRs that I created I did not see the bot interact with my PRs or anything. Do you know how it is supposed to work?

lenucksi:
Looks we're either holding it wrong or it doesn't work at least as I expected. It 's installed according to its documentation and should just show up in PRs. Take a look at the linked docs above.

dterol23:
I use in some other projects this spell checker GitHub action https://github.com/uribench/spell-check, not sure if that could also be interesting for this issue. My colleague (uribench at GitHub), wrote the spell checker and he will be happy to support the evaluation and integration if it is interesting. It can be downloaded and run locally first for testing. https://github.com/uribench/spell-check/blob/master/README.md

spier:
@lenucksi not sure what changed but now prosebot is doing stuff. I see these 3 flavors of it on some PRs now.

![Screen Shot 2021-01-17 at 10 42 57 ](https://user-images.githubusercontent.com/163029/104836832-01590900-58b1-11eb-95f9-ee332a21acdb.JPG)

spier:
[...]

PR #223, opened 2020-09-26
Fixing some typos in the MaturityModel pattern

spier:
There is at least one other typo in there ("oportunity").
Want to see if the bot catches that :)

### DIFF
--- a/patterns/2-structured/maturity-model.md
+++ b/patterns/2-structured/maturity-model.md
@@ -75,12 +75,12 @@ In order to motivate colleagues to contribute work outside of their core team th
 
 **Helpful Resources**
 
-In order to attract contributors helpful supporing material needs to be easily accessible.
+In order to attract contributors helpful supporting material needs to be easily accessible.
 
 * RS-0: Individuals and teams neither contribute to nor draw upon a shared repository of knowledge.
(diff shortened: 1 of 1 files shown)


spier:
@lenucksi how can I be sure that prosebot actually checked the .md files in this PR?

lenucksi:
> @lenucksi how can I be sure that prosebot actually checked the .md files in this PR?

From what's in the documentation it should have been visible as a pull request check here and added custom notes. 
Apparently that did not work out as expected, which means that there's the test I wanted in #218. Looks like that one needs fixing.

 Also, Linkbot was unhappy about this link: https://github.com/open-organization/open-org-distributed-work-guide/blob/master/drostfromm_remote-first-through-openess.md#setting-cross-team-standards-and-best-practices-in-the-open

The file is still present and hence but changes so either we reference the [file without anchor](https://github.com/open-organization/open-org-distributed-work-guide/blob/master/drostfromm-remote-first-through-openess.md) or just update the link

spier:
I have already fixed the broken link as part of https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/225, which has been merged to master in the meantime.

Do you want to keep this PR here open as a test for your prosebot? Otherwise I can rebase this on master and we can get it merged.

lenucksi:
Lets merge this for now and keep an intentionally defect PR around to test linters. Found some other bots / linter integrations in case prosebot doesn't go where want it to.

PR #225, opened 2020-09-26
Moving both patterns related to 'Overcome Acquisition-based Silos' 1 maturity level down

spier:
Moving both patterns related to 'Overcome Acquisition-based Silos' 1 maturity level down.

Reasoning:
- the pattern template isn't fully filled out
- doesn't have a clear Known Instance yet
- contains multiple placeholders indicating that further work is needed.


NewMexicoKid:
Looks like the Link Check is failing on the README.md? I'm okay with demoting these patterns to an initial stage and agree that the patterns need further work.

spier:
Good one @NewMexicoKid. The link checker was spot on this time around, but my last commit hopefully fixed that broken link.

spier:
The remaining two links to `https://www.fearlesschangepatterns.com/` that are broken are a mystery to me. That site is up but every once it a while it fails the link checker. 

So this PR should still be good to be merged.

lenucksi:
> The remaining two links to `https://www.fearlesschangepatterns.com/` that are broken are a mystery to me. That site is up but every once it a while it fails the link checker.
> 
> So this PR should still be good to be merged.

Looks like a timeout, SSL negotiation or something else then. Might be worth to report this behavior upstream. And / or exclude the domain from link-checking with a notice in the exclusion file for the reason. Or else we'll have spurious red checks which is going against the point of automated checks.

PR #228, opened 2020-09-29
Summarized report for the October Governance meeting 

lenucksi:
This is an edited and shortened version of the document prepared by @spier in https://docs.google.com/document/d/1XLdiLHriMsIOyEMw1BVZKlGREC5ZBAN1w_-BhzRoDB4/edit?usp=sharing

The recommendation in terms of length from @MaineC is _"at max half a page"_. We probably did not hit that threshold but it is slightly shorter.
I've split the content in an overview an a detail section, where the detail reflect the contributors that contributed great content and whose names should be named.

Happy to see discussions, please add them as commit suggestion in a review. I'll be able to take a look every European evening.

We're supposed to hand in something until Friday the 2nd October, hence I'll pull request the result of this pull request to our Governance repository at Friday the 2nd. 20:00 German time.

### DIFF
--- a/meta/boardreports/2020-10.md
+++ b/meta/boardreports/2020-10.md
@@ -16,7 +16,7 @@
 
 - New process proposal by @maxcapraro, discussion (@spier @mainec @newmexicokid @lenucksi) and successful adoption (initial, structured, validated, goal: faster first contribution, see contributor handbook)
 - Processing of existing content from wiki and existing PRs into publishable repo content, large scale cleanup, migration and fixing, all ongoing for the foreseeable future. 
-- Great work towards content rendering (e.g. website, initial PDFs) in progress @spier @dizquerdo (https://innersourcecommons.gitbook.io)
+- Great work towards content rendering (e.g. website, initial PDFs) in progress @spier @dizquerdo (https://innersourcecommons.gitbook.io/innersource-patterns-staging/v/book/)
 - Conference activity, more, interesting pattern activity on Slack leading to new patterns 
 - Patterns office-hours concept try out, some success, dual time-zone meeting concept successfully in use in marketing wg
 
--- a/meta/boardreports/2020-10.md
+++ b/meta/boardreports/2020-10.md
@@ -1,70 +1,35 @@
 # InnerSource Patterns WG - Report for governance meeting 2020-10
 
-**Q3/2020 | July - September** (unzutreffend deckt eigentlich alles letztem treffen ab)
+**Data from June to September**
 
 ## Mission - Briefly describe what your project actually does.
 
 - Discuss community InnerSource practices and dynamics, collect and document agreed upon best practices of how to do InnerSource - in the form of patterns
 - Continuously publish the most mature patterns as an ebook and website
 
-## Project/Community Status and Health - Sum up the status and health of your project and community in a few sentences.
+## Project/Community Status and Health - Sum up the status and health of your project and community in a few sentences
 
 - This project is seeing more activity than ~1 year ago, with 10 new contributors, some of them repeatedly, some major contributions, lots of evolutionary improvements. Also new contributions from past contributors.
 - Activity and trusted committer diversity is still too low to sustain the project in the long run on a high activity level. Low activity evolution is sustainable.
 
 ## Project Activity - Describe the overall activity in the project over the past quarter.
 
-- New process proposal by @maxcapraro, discussion (@spier @mainec @newmexicokid @lenucksi) and successful adoption (initial, structured, validated, goal faster first contribution, see contributor handbook)
-- Processing of existing content from wiki and existing PRs into publishable repo content, large scale cleanup, migration and fixing 
-- Great work towards content rendering (e.g. website, PDFs) in progress @spier @dizquerdo (https://innersourcecommons.gitbook.io)
-- Conference activity, more activity on Slack leading to new patterns 
-- patterns office-hours concept try out, some success, dual time-zone meeting concept successfully in use in marketing wg
+- New process proposal by @maxcapraro, discussion (@spier @mainec @newmexicokid @lenucksi) and successful adoption (initial, structured, validated, goal: faster first contribution, see contributor handbook)
+- Processing of existing content from wiki and existing PRs into publishable repo content, large scale cleanup, migration and fixing, all ongoing for the foreseeable future. 
+- Great work towards content rendering (e.g. website, initial PDFs) in progress @spier @dizquerdo (https://innersourcecommons.gitbook.io)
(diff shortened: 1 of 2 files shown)


meta/boardreports/2020-10.md
```diff
@@ -0,0 +1,68 @@
+# InnerSource Patterns WG - Report for governance meeting 2020-10
+
+**Data from June to September**
+
+## Mission - Briefly describe what your project actually does.
+
+- Discuss community InnerSource practices and dynamics, collect and document agreed upon best practices of how to do InnerSource - in the form of patterns
+- Continuously publish the most mature patterns as an ebook and website
```
spier, line 8: I just want to point out that the mission statement here is slightly different from what we have. in our repo README. Just pasting the original text in verbatim here for reference. 

Doesn't matter much to me but maybe we should just stick to the original?

```suggestion
- Collect and document agreed upon best practices of how to do InnerSource - in the form of patterns
- Continuously publish the most mature patterns as an ebook
```

meta/boardreports/2020-10.md
```diff
@@ -0,0 +1,68 @@
...
+## Project/Community Status and Health - Sum up the status and health of your project and community in a few sentences
+
+- This project is seeing more activity than ~1 year ago, with 10 new contributors, some of them repeatedly, some major contributions, lots of evolutionary improvements. Also new contributions from past contributors.
+- Activity and trusted committer diversity is still too low to sustain the project in the long run on a high activity level. Low activity evolution is sustainable.
+
+## Project Activity - Describe the overall activity in the project over the past quarter.
+
+- New process proposal by @maxcapraro, discussion (@spier @mainec @newmexicokid @lenucksi) and successful adoption (initial, structured, validated, goal: faster first contribution, see contributor handbook)
+- Processing of existing content from wiki and existing PRs into publishable repo content, large scale cleanup, migration and fixing, all ongoing for the foreseeable future. 
+- Great work towards content rendering (e.g. website, initial PDFs) in progress @spier @dizquerdo (https://innersourcecommons.gitbook.io)
```
spier: The board can already look at the latest version if they want. Adding that link instead.

```suggestion
- Great work towards content rendering (e.g. website, initial PDFs) in progress @spier @dizquerdo (https://innersourcecommons.gitbook.io/innersource-patterns-staging/v/book/)
```

meta/boardreports/2020-10.md
```diff
@@ -0,0 +1,68 @@
...
+- 24th of April 2020: @lenucksi
+
+## Committer Diversity
+
+**Cover committer diversity in your project. A healthy project should survive the departure of any single contributor or employer of contributors. What are your steps to make sure that people from all regions on the globe can participate in your project without having to be awake at midnight their local time?**
+
+- Focus on asynchronous collab in the #innersource-patterns channel
+- Offering of office hours on APAC and EU/Americas friendly timeslots
+- Work towards automation of administration tasks, documentation of processes
+- Have TCs in multiple time zones
```
spier, line 57: I am not aware of what steps the WG has done towards this last point?

meta/boardreports/2020-10.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Last Committer Addition
+
+**When was the last committer added to the project? A healthy project tends to add new committers regularly. The report should indicate the most recent date on which a committer was added.**
+
+- 24th of April 2020: @lenucksi
+
+## Committer Diversity
+
+**Cover committer diversity in your project. A healthy project should survive the departure of any single contributor or employer of contributors. What are your steps to make sure that people from all regions on the globe can participate in your project without having to be awake at midnight their local time?**
```
spier, line 52: This is main only real objection in terms of content:
> A healthy project should survive the departure of any single contribution

This is not the case right now.

To make it explicit, if @lenucksi were to leave right now, PRs would not get merged anymore, and the feedback process would slow down to a halt. e.g for me personally as a contributor it will mean that I would likely leave the project as well because then the progress would become frustratingly slow, and I would no longer be motivated to contribute due to the lack of visible progress.

If others also see this risk, then I think we should call out this risk more explicitly. I have seen the `Activity and trusted committer diversity is still too low` above but I think we should make a stronger point about this.

This is the make or break point for the Patterns WG right now.

spier (changes requested):
I have one major change request, as you will see in the inline feedback.

Other than this the report looks good to me.

In terms of report length, the input was:
> Those reports don't have to be long - half a page is enough

For me this means that the reports can be longer as well.
Also given the amount of questions (8 in total), and that we are reporting on more than a quarter rather than a single month, any report will end up with at least 1 page anyways. 

Therefore I don't think that we should get hung up on the length of the report. Reports in the future will likely be shorter anyways (as we then report on a single month only).

I have not checked the stats, but I am sure they are fine.

spier:
Hi @lenucksi, do you want any help with finishing up this report?

I cannot help with the PR to the Governance repository,  as I don't have access there.
But I can help with anything content-related on this PR here.

Just let me know.

meta/boardreports/2020-10.md
lenucksi: If we're playing by the book here, we need remove the mention of the new website hosted version as well since it's not in the plan.

The mentioned parts are of critical relevance for our outcome because without them no patterns would arise to put here. Hosting it as website is just as good as a PDF and a valuable add-on. I'll raise a PR to amend the Readme.

meta/boardreports/2020-10.md
lenucksi: As mentioned above, thanks for the updated location.

meta/boardreports/2020-10.md
lenucksi: The question does not for what have you done but what are your steps an approaches. This answers it.
Other than that: 
- 3 out of 4 points can be seen in action, the other will come once there's the time for it.

meta/boardreports/2020-10.md
lenucksi: The quoted sentence sets the context for the following questions. The questions have been answered.
We've added onboarding of other TCs to our plans which is what will address this.

lenucksi (commented):
Adressed review comments and merged commit suggestions.

NewMexicoKid (approved):
Looks good. Glad to see the community active again.

meta/boardreports/2020-10.md
spier: On this particular point my proposal is:
Let's agree to disagree :)

spier (approved):
Although I don't agree with some of the points in this report, I am still happy for it to go out as is. 

We will have many more chances to write new (monthly) reports, and through this we will learn which metrics matter most, and where the risks and opportunities for the working group are.

PR #229, opened 2020-09-30
Updated reference implementation (initial value, description bonus)

Michadelic:
* starts score calculation with 50 instead of 1 to give active projects with low GitHub stats (forks, stars, watches) a little advantage
* gives projects with a meaningful description (>30 chars) a small bonus of 50 to be listed higher than similar projects with no description


spier (approved):
Nothing to object with, give that this is how you do it 👍 

For me the comments in the pseudocode are equally if not more interesting than the code itself as they describe your intention.

Great stuff!

lenucksi (approved):
Thanks, LGTM

PR #230, opened 2020-10-03
Adding ISC Summit video from SAP to 'InnerSource Portal' and 'Repository Activity Score'

spier:
Adding ISC Summit video from SAP to two patterns:
- InnerSource Portal
- Repository Activity Score

This confirms the Known Instances for these patterns, and also helps other readers to understand those two pattern as well as the interplay between the two better.

@Michadelic can you confirm that you are happy for the video to be listed with the patterns?


Michadelic (approved):
Thanks for linking the video, looks good to me

PR #236, opened 2020-12-02
Add link to reference implementation (project portal open source project)

Michadelic:
SAP's implementation for the InnerSource portal pattern is now open source. This change adds a link and some context to the pattern body. Enjoy ;-)

PR #237, opened 2020-12-02
Adding screenshot of InnerSource Portal from SAPs open source project

spier:
Now that SAP has open sourced an implementation of an InnerSource Portal we have a convincing visual to add here 👍

Together with https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/236 this will make the InnerSource Portal pattern a lot better :)


spier:
@lenucksi if you happen to see this PR in time, getting this and https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/236 merged before day 2 of the ISC Summit tomorrow would be really awesome. We could use this as a great sample of how patterns help companies and then companies help the patterns improve. Full circle.

I know this is super short-term though, so don't worry if you don't see this in time.

lenucksi (approved):
LGTM

Issue #173, opened 2020-05-12
Enable all trusted committers to produce an ebook of proven patterns


spier:
Our mission statement says:
> Continuously publish the most mature patterns as an ebook

@dicortazar has produced an updated gitbook in April 2020. This is an interactive (clickable) version of our patterns: https://dicortazar.gitbook.io/inner-source-patterns/ (has been removed in the meantime)

## Possible Challenges (with the current approach)

* people cannot download the gitbook. When I heard "ebook" I was expecting an ebook/PDF that people can download. Maybe that expectation is not shared in this group?
* right now only @dicortazar can update the gitbook. That will lead to the gitbook being outdated quickly. I think it would be great if all trusted committers could update the ebook. Even better if that would happen automatically when changes are merged to this repo.
* the gitbook does not seem to be an official output artifact of the InnerSource Patterns community yet. For example it is not shared on [innersourcecommons.org/resources/books/](http://innersourcecommons.org/resources/books/) yet, which makes me think that it is sort of a beta release so far?

## Proposed Solution

* produce a PDF and publish that at [innersourcecommons.org/resources/books/](http://innersourcecommons.org/resources/books/)
* automate the production of the ebook, optimally so that it automatically refreshes when the master branch of this repo is updated
* **unclear**: What to do with gitbook? Do we want to have both (gitbook and a PDF) or only one?

## Tech

@MaineC shared how to generate a simple PDF using pandoc. The pattern `.md` files are listed explicitly to guarantee a certain order.

```
pandoc start-as-experiment.md dedicated-community-leader.md praise-participants.md project-roles/trusted-committer.md service-vs-library.md 30-day-warranty.md contracted-contributor.md crossteam-project-valuation.md common-requirements.md gig-marketplace.md discover-your-innersource.md improve-findability.md innersource-portal.md introducing-metrics-in-innersource.md junkyard-styled-innersourcing.md modular-code.md review-committee.md -o patterns.pdf
```

Looks like this could be further automated e.g. using [pandoc-action-example](https://github.com/pandoc/pandoc-action-example)

## References

* more details in [this Slack Conversation](https://innersourcecommons.slack.com/archives/C2EFRTS6A/p1587468876145100)

spier:
@dicortazar I would be really curious to hear your opinion about the challenges and proposal above, as you have all the working knowledge from producing the gitbook so far.

spier:
Some other relevant links if we want to use pandoc for this:

[...]

PR #239, opened 2020-12-05
Shortening some pattern titles to make for nicer links

spier:
Shortening some pattern titles to make for nicer links. Preparing for our book release (see #173).
I moved the long form of the titles to the Alias section where that seemed useful.


spier:
@MaineC I touched a couple of the patterns that you contributed in here. Are you happy with the changes that I made to the pattern titles?

spier:
@dicortazar I touched a pattern hat you contributed in here. Are you happy with the changes that I made to the pattern titles?

gruetter:
I like all of the new, more concise titles. Wondering if we need the _Alias_ section at all, @spier . I feel that the extra text in this section should not be necessary if there's a good patlet. Also, I would expect equally short, alternative titles in the _Alias_ section rather than "elaborations" on the existing title. Just my 0.02 $. @NewMexicoKid , WDYT?

spier:
Yeah, I understand your line of thinking about the Aliases.

The **Alias** section is something that we introduced a couple of months ago, and the only pattern where I recall using it so far is [Dedicated Community Manager](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/patterns/2-structured/dedicated-community-leader.md#alias), where it is indeed used like you are imagining.

For the 2 patterns with Aliases in this PR I decided to keep them in there because the old titles did seem to carry some semantic, and I didn't want to judge if that was fully expressed already outside of the title. 

Another tangental thought on the Title/Alias: 
Once we publish the book, the title will be part of the URL for the pattern too. Then we should try to keep the title and URL as stable as possible so that patterns can be referenced from elsewhere without risking a broken link (for that purpose other ISC resources like the Learning Path might want to point to the patterns in the book, rather than to the patterns in the repo).

spier:
@MaineC we are about to release the first version of our Patterns as a gitbook. Would be nice if we could get some of the Patterns titles, as that makes for an easier to consume table of contents in the book. 

Please review the diffs in this PR and let us know if you are ok with the shortened titles.

spier:
I am moving ahead with this change as is, to get book into a prettier shape (I think so at least :)).

The original titles of these patterns have still been conserved in the `## Alias` section, so they are not lost in case the original authors of these patterns want to use them for something.

In later validation rounds, e.g. when leveling-up the pattern from Structured to Validated, we might get rid of the Aliases then if they don't provide any additional value/content as @gruetter was explaining above.

PR #246, opened 2020-12-14
Fix failing checks & GitHub Actions

spier:
Multiple issues with the failing checks:

* failing Link Check
   * Excluding fearlesschangepatterns.com from link check (in GitHub Action).
   * fearlesschangepatterns keeps failing inconsistently, so Ii am excluding it.
* duplicate link check
   * fixed by removing `pull_request` from the workflow trigger. 
* book ToC Generation triggers even when not on master
   * fixed by removing `pull_request` from the workflow trigger. I hope it will only trigger now when we merge the PR (i.e. push to master)

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -1,7 +1,6 @@
 name: Book ToC Generation
 
 on:
-  pull_request:
   push:
     branches:
       - "master"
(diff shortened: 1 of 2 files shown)


spier:
Reduced the number of checks. Need to merge it to see if the ToC generation triggers when merged to master.

PR #247, opened 2020-12-16
Linking to innersource-portal-hygiene.md from README. 

spier:
This is some post-processing of #120 that I could not do in that PR itself.

Things done here:
* Linking to innersource-portal-hygiene.md from README. 
* some cleanup in the pattern itself the pattern itself.


spier:
Trivial change. Merging without TC review.

PR #249, opened 2020-12-23
Adding step for new TCs to get added to CODEOWNERS

spier:
Adding information to the onboarding instructions for new TCs, so that we don't forget about the `.github/CODEOWNERS` file.


spier:
Trivial change. Merging.

Issue #159, opened 2020-05-04
Validation of patterns against pattern template syntax


spier:
It could be helpful to validation patterns against the [Pattern Template](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/meta/pattern-template.md) for various reasons:

* bring existing patterns to a more consistent format, which helps us with the preparation for the creation of the ebook
* help new contributors by validating their pattern contributions, and pointing them to specifics parts of the Pattern Template that they missed

### Validation ideas

1. all mandatory headlines/sections are present
1. the top level headlines of the pattern all use `##` (h2) (i.e. no `#`)
1. no duplicate headlines/sections exist
1. headlines/sections are in a certain order
1. no headlines/sections exist that are not part of the pattern template

### Risks/Questions:

* the validations could be frustrating or discouraging for new contributors. however if we do it right, the validations could help them to understand the pattern format.
* what to do with sections where the headline is present but then there is actually no text filled into that section?

### Status

I tried out [markdownlint](https://github.com/DavidAnson/markdownlint), together with the the GitHub Action [markdown-linting-action](https://github.com/marketplace/actions/markdown-linting-action).

As many of the existing patterns contain optional headlines as well as custom headlines, we will likely have to write custom validation rules. I have already tried this out and it looks manageable. (see [experimental code](https://github.com/spier/InnerSourcePatterns/tree/linter-for-patterns/lint)).

My current suggestion would be to validate rules 1-4 above, but don't validate rule 5.

spier:
The validation results could look similar to this

```
$ markdownlint -c lint/pattern-template.yml -r lint/pattern-template.js start-as-experiment.md

start-as-experiment.md:1 PATTERN-TEMPLATE-RULE-003 Mandatory template headlines/sections [Required headline/section is missing] [Context: "Solutions"]
start-as-experiment.md:43 PATTERN-TEMPLATE-RULE-002 Standard Headlines [Allowed types are: Title, Patlet, Problem, Story, Context, Forces, Solutions, Resulting Context, Known Instances, Status, Author(s), Acknowledgements] [Context: "## Solution"]
start-as-experiment.md:92 PATTERN-TEMPLATE-RULE-002 Standard Headlines [Allowed types are: Title, Patlet, Problem, Story, Context, Forces, Solutions, Resulting Context, Known Instances, Status, Author(s), Acknowledgements] [Context: "## Related Patterns"]
[...]

PR #257, opened 2020-12-26
Custom linter to check the format of the patterns

spier:
This custom markdown linter rules checks the patterns and confirms if all mandatory sections specified in the [Pattern Template](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/meta/pattern-template.md) are present. Implements https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/159.

The checks are only running on patterns of maturity **Structured** (Level 2) and **Validated** (Level 3), but not for **Initial** (Level 1).
This allows patterns of maturity **Initial** to be merged faster, while we apply more scrutiny to the higher levels as they get published in. the book as. well.

The linter, which is executed by a GitHub Action, checks all patterns in these folders:
* `patterns/2-structured/*.md` 
* `patterns/2-structured/project-setup/*.md`
* `patterns/3-validated/*.md`

Luckily the patterns of level 2 and 3 have almost no linter issues anymore. However therefore it is hard to see what the linter would do if any issue exists. To see the linter in action see the bottom of this demo PR:
https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/168/files

FYI the old PR #168 is just kept around for demo purposes and won't be merged.

### DIFF
--- a/.github/workflows/lint-patterns.yml
+++ b/.github/workflows/lint-patterns.yml
@@ -1,0 +1,24 @@
+# from: https://github.com/marketplace/actions/markdown-linting-action
+
+name: Validate pattern syntax
+
+on:
+  push:
+    branches:
+      - "master"
+  pull_request:
+
+jobs:
+  validate-pattern-syntax:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v2
+    - name: Problem Matcher for markdownlint-cli
+      uses: xt0rted/markdownlint-problem-matcher@v1
+    - name: Lint pattern files (markdown)
+      uses: avto-dev/markdown-lint@v1
+      with:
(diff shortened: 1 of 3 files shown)


spier:
Currently the pattern syntax check is rightfully failing. 

Reason:
`patterns/2-structured/common-requirements.md` is missing required headline "Known Instances" 

I am checking with the author of that pattern if a "Known Instances" exists for that pattern. If not we would either have to move that pattern back to 1-initial or find a known instance somehow.

spier:
Note to self, `PATTERN-TEMPLATE-RULE-002` is not used right now, and likely can be removed before merging this PR.

lenucksi (approved):
Nice, cool stuff. LGTM. 
Should we need something fixed, changed or amended let's just have another PR and reference this one.

PR #258, opened 2020-12-26
Indicate correct status in all Level 2 Patterns (Structured).

spier:
After changing the review process and the maturity levels, we had not updated the status in all Level 2 Patterns to "Structured" yet. This PR fixes that.


spier:
Trivial change. Merging.

Issue #259, opened 2020-12-27
Solidify patterns in Maturity Level 2: Structured


spier:
Our `README.md` lists the patterns under [Maturity Level 2: Structured](https://github.com/InnerSourceCommons/InnerSourcePatterns#maturity-level-2-structured) in two sub-groups:

* Reviewed Patterns (proven and reviewed)
* Pattern Drafts (proven, not yet fully reviewed)

We inherited these sub-groups from the review process that we had before we introduced the 3 new maturity levels. Therefore I think it is time that we clean up these sub-groups.

## What to do?

I think we would have to do the following for each pattern in the group **Pattern Drafts (proven, not yet fully reviewed)** - find these patterns listed below:

* review the pattern against the [level 2 requirements](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/meta/contributor-handbook.md#requirements-level-2---structured)
* as a side effect of this work we might create a [Basic Style Guide](https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/171) as well, that can then be used for further reviews of patterns in this level

## And then what?

Once this is done, we can delete the two sub-groups from `README.md`.

Also note that all of these patterns that are already in a `.md` file (i.e. not open PRs or gDocs) are already published [in our book](https://innersourcecommons.gitbook.io/innersource-patterns/), irrespective of which sub-group they are in.

## Patterns that need to be reviewed

Pattern | Notes | Move Action
------------ | ------------- | -------------
**Assisted Compliance** | (**done**)  pattern in PR #74 | moved to L1 group
**[Cross-Team Project Valuation](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/patterns/2-structured/crossteam-project-valuation.md)** | (**done**) review in #261 | moved to main L2 group
**[Reluctance to Receive Contributions](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/patterns/1-initial/reluctance-to-accept-contributions.md)** | (**done**) review in #282 | moved to L1 group
**What Before How or Services Documentation** | (**done**) pattern in a Google doc (I requested access to this doc but not sure if we can recover it) | removed the pattern in #284 
**Open Source Trumps InnerSource** | (**done**) pattern in PR #46  | moved to L1 group
[...]

PR #260, opened 2020-12-27
Fix gitbook formatting issues in 3 patterns from Europace

spier:
This work is part of #259. 

Changes in overview:
- remove line-breaks in lists, as otherwise the list items don't use the full width of the book pages. extra line-breaks in lists may also lead to broken links
- link to other patterns where possible
- format file names with `filename` syntax
- introducing sub-sections for both `README.md` and `CONTRIBUTING.md`
- further minor typos and formatting inconsistencies


spier:
@MaineC as these are mostly formatting changes that will immediately improve the layout of your patterns in our gitbook, I will go ahead and merge this without your review.

However feel free to send updates in a separate PR, should I have changes something in a way that you don't like.

PR #261, opened 2020-12-27
Review crossteam-project-valuation

spier:
Part of #259. 

The changes so far are mostly:
- Formatting fixes in the markdown. 
- Changes to spelling of InnerSource.

I have some further questions for @rrrutledge that I will put into inline comments.


patterns/2-structured/crossteam-project-valuation.md
```diff
@@ -37,18 +37,19 @@ There is a common process, though, by which you can create a model to calculate
...
 The idea is for the group of experts to be able to say to each other, "We don't know exactly how long it would take, but we can all agree it's _at least_ this much."
 Specifically, you should estimate a *maximum* reasonable time to consume your project output and *minimum* reasonable times for consumers to otherwise home-roll, use and maintain their own solutions.
 
-> One note about cost-to home-roll.  The cost to home-roll a solution is NOT necessarily (very unlikely, in fact) the same as the cost of making a shared solution.
-Oftentimes for the same functionality the modularity and quality involved in buliding a cross-team, shared solution makes it a noticeably higher investment than a quick, hardcoded implementation used just once.
+> One note about cost to home-roll.  The cost to home-roll a solution is NOT necessarily (very unlikely, in fact) the same as the cost of making a shared solution.
+> Oftentimes for the same functionality the modularity and quality involved in building a cross-team, shared solution makes it a noticeably higher investment than a quick, hardcoded implementation used just once.
```
spier: Two questions about this section:
- what does "home-roll" mean? Is that similar to "roll your own (solution)"?
- why is this section here in block-quotes? Could it be equally well be part of the regular text?

patterns/2-structured/crossteam-project-valuation.md
```diff
@@ -37,18 +37,19 @@ There is a common process, though, by which you can create a model to calculate
...
 
 ## Formula
```
spier: This pattern contains 3 extra sections that are not part of the regular pattern template format:
- Explanation
- Formula
- Commentary

Some thoughts:
- I wonder if these should be considered sub-sections to "Solution"? (in which case we could move them one heading level down i.e. to `###`. This would likely be the quickest fix to make the format of this pattern more consistent with other.s
- Alternatively we could work the content from the 3 extra sections into other sections?
- **Explanation** could maybe be renamed to **Rationale**?

patterns/2-structured/crossteam-project-valuation.md
lenucksi: Reads like "roll your own" to me. Also, don't see any reason for block quote.
```suggestion
One note about cost of "rolling your own solution" (home-roll).  The cost to home-roll a solution is NOT necessarily (very unlikely, in fact) the same as the cost of making a shared solution.
Oftentimes for the same functionality the modularity and quality involved in building a cross-team, shared solution makes it a noticeably higher investment than a quick, hardcoded implementation used just once.
```

lenucksi (approved):
Interesting changes. Let's get the discussions resolved and merge it.

patterns/2-structured/crossteam-project-valuation.md
spier: Agreed!

spier:
(moved inline comment)

This pattern contains 3 extra sections that are not part of the regular pattern template format:
- Explanation
- Formula
- Commentary

Some thoughts:
- I wonder if these should be considered sub-sections to "Solution"? (in which case we could move them one heading level down i.e. to `###`. This would likely be the quickest fix to make the format of this pattern more consistent with other.s
- Alternatively we could work the content from the 3 extra sections into other sections?
- **Explanation** could maybe be renamed to **Rationale**?

patterns/2-structured/crossteam-project-valuation.md
spier: Moving this comment to the main comment thread.

spier:
I went for the least invasive chance which was to move the 3 sections Explanation/Formula/Commentary as sub-sections under 'Solution'.

@rrrutledge if you could review these changes to check that we didn't misrepresent anything here, that would be awesome.

As these are mostly formatting changes, typo fixes, and minimal changes to make the pattern fit he pattern template, I would otherwise merge this PR towards the end of next week or so.

spier:
Reaching self-set timer of this PR :). Merging.

PR #263, opened 2020-12-28
fix: MD010/no-hard-tabs

nschonni:


spier:
Wow @nschonni, that was a quick follow-up PR. Nice! :)

Could you please add a bit more description to this PR to help me understand how you are going about these formatting fixes? 

e.g. some questions that I have are:
- is the end goal to have all linter rules switched on again?
- how did you decide to switch on `MD010` first? or just random?
- I suspect you used `markdownlint -f "**/*.md"` to fix the issue reported by `MD010`? I tried this locally and now I get a bunch of error reports about "_MD005/list-indent Inconsistent indentation for list items at the same level_"

Would be great to learn more about this. 

Cheers.

nschonni:
>     * is the end goal to have all linter rules switched on again?

Probably most, but not all things like MD013 (line length) make sense for all projects


>     * how did you decide to switch on `MD010` first? or just random?

Random, but usually a non-controversial change unless tabs are being used everywhere

>     * I tried this locally and now I get a bunch of error reports about "_MD005/list-indent Inconsistent indentation for list items at the same level_"

Yes, autofixed first, then manually corrected afterwards

meta/contributor-handbook.md
```diff
@@ -26,37 +26,37 @@ To achieve a given maturity level a pattern has to satisfy the requirements for
...
 ### Requirements: Level 2 - Structured
 
 Patterns of level 2 are stored in the directory `/patterns/2-structured`.
 
 - Validation requirements:
-	- Is validated by at least one known instance 
-		- Alternatively: key elements of the pattern have been validated in separate contexts and, in consequence, it is justified to believe the full solution will function
-	
+  - Is validated by at least one known instance 
+  - Alternatively: key elements of the pattern have been validated in separate contexts and, in consequence, it is justified to believe the full solution will function
```
spier: I checked the visual diff on this file, and the bullet point starting with "Alternatively" was actually meant to be one level lower. 

Therefore we should change it to:

```suggestion
  - Is validated by at least one known instance 
    - Alternatively: key elements of the pattern have been validated in separate contexts and, in consequence, it is justified to believe the full solution will function
```

spier (changes requested):
Looks good generally. Requested one change.

spier:
Thanks for one more step towards cleaner markdown @nschonni ! 👍

PR #264, opened 2020-12-30
fix: Markdownlint rules for whitespace

nschonni:
Fixes the Markdownlint rules related to spacing around headings, lists, and trailing whitespace.
There are still some trailing double spaces it doesn't remove, because those can be intentional newlines in MD (but in likely in most cases here)
This doesn't show any changes if you look at the files in the "Smart diff"


lenucksi (approved):
LGTM

PR #268, opened 2021-01-07
Typos fixed in maturity model pattern

dterol23:
While reading the maturity model pattern, I found a couple of typos which I propose to fix. 
I also added a general comment about spell check CI in #218


spier (approved):

spier:
Thanks for these fixes @dterol23!

They are already visible in the published pattern in the book now!
https://innersourcecommons.gitbook.io/innersource-patterns/patterns/maturity-model

PR #270, opened 2021-01-09
Link to book from README

spier:
I don't actually know how people find this repo, but I suspect some come via Google searches.

For those of them that are the "learning by reading" type, I suspect that the book would be the better starting point for them. Therefore I want to make it easier for them to find the book.

What I changed:
- adds the book cover image to the README
- links to the book immediate at the top of the README (right after explaining what patterns are in the first place)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,16 @@ You are already using InnerSource in your company and want to share your ideas a
 
 ## Mission Statement
 
+<img align="right" src="book/innersource-patterns-book-cover.jpg" width="50%">
+
 Our mission in this working group
 
 - Collect and document agreed upon best practices of how to do InnerSource - in the form of patterns
 - Continuously publish the most mature patterns as an ebook
+
+## The book
+
+We publish the most mature InnerSource Patterns at [innersourcecommons.gitbook.io/innersource-patterns](https://innersourcecommons.gitbook.io/innersource-patterns/). With "most mature" in this context we mean level 2+3 patterns, as shown below.
 
 ## List of Patterns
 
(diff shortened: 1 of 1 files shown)


spier:
@fwan2000 I think you can related to the "learning by reading" users very well. What do you think about this change?

README.md
```diff
@@ -1,7 +1,12 @@
 # InnerSource Patterns
 
+<a href="https://innersourcecommons.gitbook.io/innersource-patterns/">
+<img align="right" src="book/innersource-patterns-book-cover.jpg" title="The InnerSource Patterns book" width="30%"></a>
```
lenucksi, line 4: Wasn't there a way to do that with Markdown too?

lenucksi (approved):
- Looks fine to me, added one comment.

- We should also add our patterns videos on YT for the people that like videos, there are actually two rather popular videos explaining the "what, why, how" aspects in our YT account.

- @spier There's some traffic stats here: https://github.com/InnerSourceCommons/InnerSourcePatterns/graphs/traffic

README.md
spier: I think one can do simple things like a clickable image that then leads to a URL. However I am not aware of a markdown construct that would allow us to do the `align="right"` and `width="30%"`.

fwan2000:
> @fwan2000 I think you can related to the "learning by reading" users very well. What do you think about this change?

I like this change. It's easier to consume the patterns from the book. Maybe also add something immediately after that to mention come to repo for any contribution?

spier:
Thanks for the feedback @fwan2000. I have opened a new discussion in https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/275 to define a bit more clearly how we want to treat the different resources related to the InnerSource Patterns.

PR #271, opened 2021-01-09
Reduce unneeded Pattern Syntax check executions

spier:
Improves upon #257.

Limits the executions of this check to PRs that touch Level 2+3 patterns or the config of the Pattern Syntax linter itself
Always runs on `master`, just to be safe.


spier:
Trivial change. Merging.

PR #272, opened 2021-01-09
[WIP] Spell check action

spier:
Experiment with spell checker action, related to #218


spier:
I am trying out this action at the moment [reitermarkus/aspell](https://github.com/reitermarkus/aspell).
There is also this one, which already seems to have some following: https://github.com/marketplace/actions/github-spellcheck-action

lenucksi:
> 
> 
> I am trying out this action at the moment [reitermarkus/aspell](https://github.com/reitermarkus/aspell).
> There is also this one, which already seems to have some following: https://github.com/marketplace/actions/github-spellcheck-action

Looks interesting, but can it also do PR annotation? I mean, I'll look into the CI logs, but not everyone will I guess.

spier:
Yeah good question. I am playing with two different Actions here, and the 1st one shows inline annotations, for the 2nd one I couldn't figure it out yet.

nschonni:
I tend to use https://github.com/streetsidesoftware/cspell since it has good support in VS Code and CLI for CI. There is an Action, but I tend just to use the CLI directly

spier:
> as good support in VS Code and CLI for CI. There is an Action, but I tend just to use the CLI directly

Thanks for the pointer. 

From the first tests I don't see a clear path yet for how to make the spell checker functional across all of our text.
- company names, people names
- British English vs American English
- colloquialism 
- random stuff (I would argue mistakes of the spell checkers)

Will do some more test.

Just like for the markdown linter it feels like it will be important in the configuration of the spell-checking to find a sensible balance between being to strict and too lax.

nschonni:
> Looks interesting, but can it also do PR annotation? I mean, I'll look into the CI logs, but not everyone will I guess.

FYI, you can also create your own problem matchers when jobs don't exist for a particular tool. EX: https://github.com/w3c/aria-practices/blob/master/.github/workflows/cspell-problem-matcher.json and then register it https://github.com/w3c/aria-practices/blob/master/.github/workflows/spelling.yml#L44

spier:
#519 is superseding this PR here.

PR #280, opened 2021-01-18
'Pattern of the Month' approach

spier:
I am proposing to start an experiment together with the Marketing Working group in order to create some engagement around specific patterns in our InnerSource Patterns collection.

The idea is to  pick a **Pattern of the Month** and do some specific activities and advertizing around it.

More details of the idea in the is PR.

Looking forward to all comments about this 🎉 🎈

### DIFF
--- a/meta/pattern-of-the-month.md
+++ b/meta/pattern-of-the-month.md
@@ -1,8 +1,8 @@
 # Pattern of the Month
 
-We are constantly improving the [InnerSourcePatterns repo](https://github.com/InnerSourceCommons/InnerSourcePatterns), by asking the community to contribute their experience t0 existing patterns, as well as contributing entirely new patterns.
+We are constantly improving the [InnerSourcePatterns repo](https://github.com/InnerSourceCommons/InnerSourcePatterns), by asking the community to contribute their experience to existing patterns, and create entirely new patterns.
 
-However with more than 40 patterns in various maturity levels in our repo, it can be hard for new people to figure out where to get started.
+However with 40+ patterns in various maturity levels, it can be hard for new people to figure out where to get started.
 
 With the **‌Pattern of the Month** we are putting the focus on one of the existing pattern at a time, allowing the community to learn more about this one pattern together.
 
--- a/meta/pattern-of-the-month.md
+++ b/meta/pattern-of-the-month.md
@@ -1,0 +1,33 @@
+# Pattern of the Month
+
+We are constantly improving the [InnerSourcePatterns repo](https://github.com/InnerSourceCommons/InnerSourcePatterns), by asking the community to contribute their experience t0 existing patterns, as well as contributing entirely new patterns.
+
+However with more than 40 patterns in various maturity levels in our repo, it can be hard for new people to figure out where to get started.
+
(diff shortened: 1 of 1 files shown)


NewMexicoKid (approved):
Aside from the grammatical typos the prosebot is catching, the idea sounds pretty good. Is the thought to have the conversation (for it ideally should be a conversation) in the #innersource-patterns chatroom? Or would it be useful to schedule a 40 minute WebEx to have the conversation live? 
Previously, we had some very fruitful live discussions on patterns. It might be worthwhile to have some people who know the pattern well (the authors and contributors) on such a call to help describe nuances of the original pattern and to help those on the call determine what additional tweaks might be needed to make the pattern work in their context.
Asynchronous chat can also be helpful, but not everyone is comfortable or catches such chat.
Other open source communities have somehow managed e-mail threads; this hasn't been the pattern for the ISC crowd (not sure why).

spier:
That's great input @NewMexicoKid! I haven't thought through the type of events in which we would engage with the community. Some type of live discussion sounds great for sure.

I also want to find ways to make async discussions in #innersource-patterns Slack or elsewhere work. We can just try various formats and see what works.

I have adapted the proposal and added:

* We moderate discussions about the pattern:
    * **Q&A sessions** (async) in #innersource-patterns (Slack)
    * **Video conferences** (live) with people who know the pattern well (the authors and contributors) to help describe nuances of the pattern and to help the audience determine what additional tweaks might make the pattern work in their context.

claredillon:
Love the ideas above. We can promote the pattern of the month with any call to actions (e.g. join the async discussions or the discussion call) on Twitter and LinkedIn. It might be worthwhile to create a graphic background/social card for the "pattern of the month". Think of the way we have a standard graphic template for summit talks...

fwan2000:
Great idea @spier ! I think this can help people getting more value from InnerSource patterns and can be a great community building event. I like @NewMexicoKid 's suggestion of having a live session to "help describe nuances of the original pattern and to help those on the call determine what additional tweaks might be needed to make the pattern work in their context". That could help more people have better understanding of a pattern and more confidence applying it to their work. Seems to be a good way to get the community together. 

I also like the two pattern candidates for this experiment. Both are interesting to anyone doing InnerSource.

Thanks for leading this @spier !

Michadelic:
Reads well and absolutely makes sense as proposed. I suggest to create some kind of template image for this initiative or for the promoted pattern to efficiently share on social media and catch attention.

spier:
Good feedback from the Marketing Working Group:

The first time we try this out we might call this: **Pattern Deep Dive** instead, to not set the expectations that we will do this monthly right away. If the concept takes off, and we have enough people to facilitate the process then we can eventually do this monthly. But we don’t have to :)

spier:
For a while we have been publishing at least 1 pattern in the [monthly Marketing newsletter.](https://us1.campaign-archive.com/home/?u=0363fb06e6a59d48c6e260f1b&id=97224b5ff2) So there is already a process in place to do this, and also a list (somewhere) of all patterns that have already been published in the newsletter.

Therefore I think that this PR here is outdated, and will hence close it.

PR #282, opened 2021-01-20
update README.md for reluctance-to-accept-contributions category

fwan2000:
https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/patterns/1-initial/reluctance-to-accept-contributions.md . This pattern is a summary pattern to introduce four other related patterns. It lays out the common scenarios why a contribution is not welcomed. But only one other pattern is well defined (30 days warranty). Other threes are unknown. Update README.md for the pattern to be initial.


spier:
Part of #259

spier:
Thanks for the mini fix @fwan2000 . Merging!

PR #286, opened 2021-02-02
Introducing Hall-of-Fame for former TCs

spier:
While a Trusted Committer (TC) is in principle appointed for lifetime, it does happen that the interests or priorities of a TC change and they don't have enough time any more to contribute to the project.

In those cases we ask them if we should move them to our Hall of Fame. Doing so allows us to appropriately thank them for all of their fantastic contributions. Of course they can always start contributing again in the future and go back to being Trusted Committers.

Maintaining a list of TCs with a minimum level of activity has some advantages:

* **for the community** - more clarity from whom they get expect feedback for their PRs, and who can answer questions about the project
* **for the inactive TCs** - by removing them from the `CODEOWNERS` file we reduce the GitHub notifications and emails that they receive from us that they likely don't need

Thank you @nyeates or @gruetter for all your fantastic contributions to the InnerSource Patterns over the years. Of course we know that you love the Patterns dearly, so hopefully you will still contribute new thoughts to this community as you discover new InnerSource Patterns in the wild :)

I will leave this open **until Feb 9th** to see if @nyeates or @gruetter object to this. If they do, and if they would rather remain in the TCs list, we are more than happy to keep them there. (I have already partly confirmed this in private Slack conversations but I wanted to put this into the open as well to make the process more transparent).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,7 +8,7 @@
 # the repo. Unless a later match takes precedence,
 # @global-owner1 and @global-owner2 will be requested for
 # review when someone opens a pull request.
-*       @lenucksi @nyeates @gruetter @NewMexicoKid @cewilliams @spier
+*       @lenucksi @NewMexicoKid @cewilliams @spier
 
 # Order is important; the last matching pattern takes the most
 # precedence. When someone opens a pull request that only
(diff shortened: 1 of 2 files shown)


gruetter:
No objection, you honour and +1 for handling it this way.

spier:
Timer reached (Feb 9th).
Merging this in.

As this is the first time we are reducing the group of Trusted Committers, let's revisit the concept in the future again and adapt it to the changing maintenance needs of our community as we see fit.

PR #106, opened 2018-09-26
Pattern describing enterprise InnerSource Guidance Group

ttyler01:
A company may require a mechanism to set broad governance and behavioral guidelines as part of the rollout and sustaining needs of an InnerSource initiative.

PR #289, opened 2021-02-06
Incorporate InnerSource-Guideance-Group as an initial pattern

spier:
This new branch and PR is a continuation of #106.

The goal is to get the **InnerSource Guideance Group** Patten incorporated into the repo as as an initial pattern as quickly as possible, so that we can share it more easily with the wider ISC community. I wasn't sure if the author of the original PR is still available, hence I opted for this approach.


spier:
@ttyler01 I moved your pattern over here, as that will make it easier for me to help you get this merged.

I am especially curious if you have made more experiences with this approach at your work place that we could use for **Known Instances**, and **Resulting Context**?

spier:
I am opting to merge this as is, as the previous PR #106 that this is based on was so old, that I am not sure if we can reach the original author any more. 

By merging this we are at least putting the pattern into the open, increasing the chance of more people to read it and contribute to it.

spier:
@ttyler01 I opted to merge this PR already, to make it easier for more people to read your pattern.

Of course we would still be excited to hear more from you, if you have learned more about this topic in the meantime that you can share.

PR #290, opened 2021-02-07
Removing old status flags + Various non-content cleanup

spier:
We changed the maturity model some time back. About 1 year ago I believe.
We have now transitioned all patterns in the correct maturity levels and folders in this repo.

Some patterns still contained traces of the old review process, which isn't helpful any more.

This PR is cleaning up these things:

- **Important**: No content changes.
- Removing old status 'Proven' from most **Structured** Patterns
- Removing old status 'Unproven', 'Draft' etc from **Initial** Patterns
- Replacing other placeholder text with 'TBD'
- Removing empty sections that aren't required
- Change section order in some patterns
- Some typo fixes
- Some smaller formatting fixes


spier:
I doubt that any of this is controversial, hence merging without review.

PR #95, opened 2018-03-30
Update modular-code.md

ErinMB:
Erin, Tim & Keerthi revised various sections. Team - please review!

PR #293, opened 2021-02-13
Expand on patlet/problem of Modular Code Pattern

spier:
After merging #95 I noticed that the summary in the [README](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/README.md) does not fit to the problem statement in the pattern.

Summary in README.md:
> Management does not want to spend the extra resources needed to develop modular components and make them available in a visible repository for others to use.

Problem statement in pattern:
> Development does not want to spend the extra resources needed to develop modular components. As a result, shared components are not reused.

## What now?

To get both in-sync, I started to elaborate on both patlet and problem section of this pattern.
If anybody could help here, that would be awesome.

## TODO

- [x] elaborate on Problem statement in the pattern
- [x] add a Patlet to the pattern
- [ ] once we are happy with this PR, we should copy the content of Patlet to the README file as well

### DIFF
--- a/patterns/1-initial/modular-code.md
+++ b/patterns/1-initial/modular-code.md
@@ -39,7 +39,7 @@ Engineering does not want to spend the extra resources needed to develop modular
 * Provide incentives to teams to invest in modular code. Modular code is far more reusable. This could work well for large teams when working on modularized projects; team members can focus on their smaller assigned tasks.
     - Developers could get an opportunity to increase their influence in the organization.
     - Modular code makes the pathway to open source smoother.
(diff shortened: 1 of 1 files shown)


spier:
@NewMexicoKid I was wondering if you could help to review this, as you were part of the original authors?

NewMexicoKid (approved):
Suggested change looks fine (different organizations might have management or developers in a responsible role as mentioned in the patlets). Engineering is a nice phrase that encompasses both.

spier:
Note to self @spier :
Add the new Patlet to the overview in the Readme.

spier:
> Note to self @spier :
> Add the new Patlet to the overview in the Readme. 

Done

PR #300, opened 2021-02-27
Fix display of lists in 30 Day Warranty pattern

spier:
Fix display of lists in 30 Day Warranty pattern, so that they get displayed correctly in the book


spier:
Merging without review, as this contains no content change.

PR #301, opened 2021-02-27
Fix contracted contributor book display

spier:
Changes:

* Removing extra linebreaks, to make lists get displayed correctly in the book.
* While at it, also fixed some typos


spier:
No semantic changes, so merging without review.

PR #303, opened 2021-02-27
Rename dedicated community manager

spier:
The pattern in question is currently called [Dedicated Community Manager](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/patterns/2-structured/dedicated-community-leader.md), however that may have been a mistake when choosing between the available options for the pattern title (see bdefaafd10bb2a2c883d4da3dca4268fb7dd725b).

The more appropriate name for the pattern might be **Dedicated Community Leader**, for various reasons:
* other patterns are already referring to it as Leader e.g. [Contracted Contributor](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/master/patterns/2-structured/contracted-contributor.md)
* the file name is already `dedicated-community-leader.md`, so in the past there was likely a debate about the title `Leader` and `Manager` already

### DIFF
--- a/patterns/1-initial/incentive-mechanisms-for-voluntary-contribution.md
+++ b/patterns/1-initial/incentive-mechanisms-for-voluntary-contribution.md
@@ -30,7 +30,7 @@ InnerSource project.
 
 As a result, the total number of contributors remained restricted to the core team and the
 project cannot build a community of developers. Furthermore, contributions mostly originated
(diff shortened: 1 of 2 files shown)


spier:
@gruetter I think you were involved when we discussed whether to call this `Manager` or `Leader`. Would you give 👍  to this change?

spier:
There was no feedback on this, so I am moving ahead with the title change of the pattern to **Dedicated Community Leader**.

/cc @gruetter

PR #309, opened 2021-03-03
Update junkyard-styled-innersourcing.md

arlou:


spier:
Thanks for the fix @arlou !

PR #320, opened 2021-03-24
Fix broken link to InnerSource Checklist book

spier:
With the new website, some URLs from the old site no longer work.
Apparently only one such URL was used in the patterns, which this PR fixes.


spier:
Non controversial change. Merging without review.

Issue #224, opened 2020-09-26
Rename master to main (early 2021)


spier:
Many communities are considering renaming the default branch name of their repository from `master` to something else (e.g. `main`).

GitHub recommends to "waiting until later this year (2020)" with the renaming, as they are "investing in tools to make renaming the default branch of an existing repository a seamless experience for both maintainers and contributors". Also see [further info](https://github.com/github/renaming).

I am putting this here as a reminder for us as the friends of InnerSource Patterns to keep this change in mind for later this year :)

spier:
Latest [update from GitHub](https://github.com/github/renaming#creating-a-seamless-move-for-existing-repositories-) about this:
> Beginning in January 2021, we'll begin rolling out functionality to make it easier to rename the default branch of an existing repository

So I think we can wait for that, and then change our default branch.

spier:
We now have the option to do this with some support by built-in GitHub features. See screenshot:
![Screen Shot 2021-04-23 at 21 01 49 ](https://user-images.githubusercontent.com/163029/115918077-5770b600-a477-11eb-9ad7-0aa5a1edc7bc.JPG)

Things to consider:

* **people with access to this repo** (that are very few) have to change their local clones as shown in the screenshot
* **people with forks** may want to get rid of the `master` branch on their forks too. If they fetch/pull from upstream again then they would get the new `main` branch automatically
* **GitHub Action** We would also have to updated some of our [GitHub Action workflows](https://github.com/InnerSourceCommons/InnerSourcePatterns/tree/master/.github/workflows), as some use `master` as a trigger.

spier:
Will keep this open for a bit to do two things:

1. work in any feedback we receive from people with clones/forks of our repo
1. see what the gitbook Support team comes back with about the [issue with showing 2 variants](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/324#issuecomment-830593138) in the gitbook

spier:
With help from the gitbook support team I got rid of the 2nd variant published.
So now the only one left is the one published from `main`, which is what we want.

spier:
I am opting to close this issue. Might reopen if we get any unexpected feedback from people with clones/forks of this repo.

PR #324, opened 2021-05-01
Make the switch from master to main branch

spier:
This implements #224 .

## TODO

- [x] rename `master` to `main` via the GitHub project settings
- [x] update GitHub Actions to trigger on `main`
- [x] change gitbook integration to build from `main`
- [x] inform all people with forks and clones that they have to make the switch too

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -3,7 +3,7 @@ name: Book ToC Generation
 on:
   push:
     branches:
-      - "master"
+      - "main"
       - "book-staging"
 
 jobs:
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -6,7 +6,7 @@ name: Link Check on Patterns and README
 on:
   push:
     branches:
-      - "master"
+      - "main"
   pull_request:
 
 jobs:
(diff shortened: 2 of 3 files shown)


spier:
By default, gitbook assumes that you want to sync from is `master`.
Once I configured in gitbook that I want to sync `main` instead, it started showing 2 variants/versions in our book, as shown in the screenshot.

Screen Shot 2021-05-01 at 11 10 19

Not a problem for the users really, as the URLs of the pattern in the book don't change.
However it is a bit annoying, as we don't actually have 2 different variants of the book.

I have raised this question with the gitbook Support team. Does not stop us from moving ahead with this issue though.

 Docs:
 https://docs.gitbook.com/editing-content/variants

spier:
FYI @voborgus I learned today that [renaming the branch via GitHub directly](https://github.com/github/renaming#rename-existing) will do this:
> Redirect web requests for the old branch name to the new branch name

So we don't have be concerned with external pages linking directly to patterns files on the older `master` branch, as they will be redirected correctly.

That's a handy feature!

spier:
Merging to make sure that GitHub actions still run on all future PRs/pushes.

PR #328, opened 2021-05-26
Template for future Board reports

spier:
I added a template for Board reports of the Patterns working group.

The goal of this is to:
a) make it easier for others to write these report
b) by suggesting a specific format, keeping the report fairly short (~half a page)

I hope that the formatting will keep it short based on this reasoning:
- only 3 mandatory blocks in total
- the first block with the Summary contain key facts in a table format (which makes it unlikely for people to put a lot of text in there)


spier:
I think that this is a good start.

Will merge this PR and then test the template the next time somebody else from the Patterns WG creates a Board report.

PR #331, opened 2021-06-13
Run link checker action daily

spier:
Run link checker action daily, so that we catch broken links quickly.
Especially important to catch and fix links in the book (Patterns Level 2+3).

Also see: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -10,7 +10,7 @@ on:
   pull_request:
   schedule:
     # * is a special character in YAML so you have to quote this string
-    - cron:  '0 8 * * *'
+    - cron:  '30 8 * * *'
 
 jobs:
   linkChecker:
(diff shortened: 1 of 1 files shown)


spier:
While reading more about link checking in general, I noticed that [liche](https://github.com/raviqqe/liche), the tool that we are using currently, is deprecated.

I looked into switching to [lychee](https://github.com/lycheeverse/lychee), however lychee does not support link checking for links to local files yet, which is essential for us as our patterns often reference other patterns that way. See also https://github.com/lycheeverse/lychee/issues/21

spier:
What I am hoping is that now our link checker will automatically run at 8:30 UTC every day.

The link checker should alert us of any broken links in our patterns.
I assume that everybody with WRITE access to this repo will receive those alerts but I am not sure about that.

PR #341, opened 2021-08-13
Update activity score calculation algorithm

3cpt:
This PR contains changes to the [Repository Activity Score](https://patterns.innersourcecommons.org/p/repository-activity-score) algorithm.

* replace `watchers_count` by `subscribers_count` - because stars are the same as watchers, so I just kept stargazers and replaced watchers by subscribers (read more about this [here](https://docs.github.com/en/rest/reference/activity#starring-vs-watching)).
* fix slice number - I'm not a JS pro, but seems that we just want to slice the 13 (~3 months)
* simplify check - some validations can be made by just add the `?`
* repo JSON access standardization - JS versatility allow using `json['prop']` or json.prop. I picked one for better reading.

### DIFF
--- a/patterns/2-structured/repository-activity-score.md
+++ b/patterns/2-structured/repository-activity-score.md
@@ -65,14 +65,14 @@ function calculateScore(repo) {
 
     // updated in last 3 months: adds a bonus multiplier between 0..1 to overall score (1 = updated today, 0 = updated more than 100 days ago)
     let iDaysSinceLastUpdate = (new Date().getTime() - new Date(repo.updated_at).getTime()) / 1000 / 86400;
-    iScore = iScore * ((100 - Math.min(iDaysSinceLastUpdate, 100)) / 100);
+    iScore = iScore * ((1 + (100 - Math.min(iDaysSinceLastUpdate, 100))) / 100);
 
     // evaluate participation stats for the previous  3 months
     repo._InnerSourceMetadata = repo._InnerSourceMetadata || {};
     if (repo._InnerSourceMetadata.participation) {
         // average commits: adds a bonus multiplier between 0..1 to overall score (1 = >10 commits per week, 0 = less than 3 commits per week)
         let iAverageCommitsPerWeek = repo._InnerSourceMetadata.participation.slice(repo._InnerSourceMetadata.participation.length - 13).reduce((a, b) => a + b) / 13;
-        iScore = iScore * ((Math.min(Math.max(iAverageCommitsPerWeek - 3, 0), 7)) / 7);
+        iScore = iScore * ((1 + (Math.min(Math.max(iAverageCommitsPerWeek - 3, 0), 7))) / 7);
     }
 
     // boost calculation:
--- a/patterns/2-structured/repository-activity-score.md
+++ b/patterns/2-structured/repository-activity-score.md
@@ -58,7 +58,10 @@ function calculateScore(repo) {
     // initial score is 50 to give active repos with low GitHub KPIs (forks, watchers, stars) a better starting point
     let iScore = 50;
     // weighting: forks and watches count most, then stars, add some little score for open issues, too
-    iScore += repo.forks_count * 5 + (repo.subscribers_count ? repo.subscribers_count : 0) + repo.stargazers_count / 3 + repo.open_issues_count / 5;
+    iScore += repo.forks_count * 5;
+    iScore += (repo.subscribers_count ? repo.subscribers_count : 0);
+    iScore += repo.stargazers_count / 3;
+    iScore += repo.open_issues_count / 5;
     let iDaysSinceLastUpdate = (new Date().getTime() - new Date(repo.updated_at).getTime()) / 1000 / 86400;
     // updated in last 3 months: adds a bonus multiplier between 0..1 to overall score (1 = updated today, 0 = updated more than 100 days ago)
     iScore = iScore * (1 + (100 - Math.min(iDaysSinceLastUpdate, 100)) / 100);
--- a/patterns/2-structured/repository-activity-score.md
+++ b/patterns/2-structured/repository-activity-score.md
@@ -58,7 +58,7 @@ function calculateScore(repo) {
     // initial score is 50 to give active repos with low GitHub KPIs (forks, watchers, stars) a better starting point
     let iScore = 50;
     // weighting: forks and watches count most, then stars, add some little score for open issues, too
-    iScore += repo["forks_count"] * 5 + repo["watchers_count"] + repo["stargazers_count"] / 3 + repo["open_issues_count"] / 5;
+    iScore += repo.forks_count * 5 + (repo.subscribers_count ? repo.subscribers_count : 0) + repo.stargazers_count / 3 + repo.open_issues_count / 5;
     let iDaysSinceLastUpdate = (new Date().getTime() - new Date(repo.updated_at).getTime()) / 1000 / 86400;
     // updated in last 3 months: adds a bonus multiplier between 0..1 to overall score (1 = updated today, 0 = updated more than 100 days ago)
     iScore = iScore * (1 + (100 - Math.min(iDaysSinceLastUpdate, 100)) / 100);
@@ -66,7 +66,7 @@ function calculateScore(repo) {
     repo._InnerSourceMetadata = repo._InnerSourceMetadata || {};
     if (repo._InnerSourceMetadata.participation) {
         // average commits: adds a bonus multiplier between 0..1 to overall score (1 = >10 commits per week, 0 = less than 3 commits per week)
-        let iAverageCommitsPerWeek = repo._InnerSourceMetadata.participation.slice(repo._InnerSourceMetadata.participation - 13).reduce((a, b) => a + b) / 13;
+        let iAverageCommitsPerWeek = repo._InnerSourceMetadata.participation.slice(13).reduce((a, b) => a + b) / 13;
         iScore = iScore * (1 + (Math.min(Math.max(iAverageCommitsPerWeek - 3, 0), 7)) / 7);
     }
     // boost calculation:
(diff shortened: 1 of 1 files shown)


patterns/2-structured/repository-activity-score.md
```diff
@@ -58,15 +58,15 @@ function calculateScore(repo) {
     // initial score is 50 to give active repos with low GitHub KPIs (forks, watchers, stars) a better starting point
     let iScore = 50;
     // weighting: forks and watches count most, then stars, add some little score for open issues, too
-    iScore += repo["forks_count"] * 5 + repo["watchers_count"] + repo["stargazers_count"] / 3 + repo["open_issues_count"] / 5;
+    iScore += repo.forks_count * 5 + (repo.subscribers_count ? repo.subscribers_count : 0) + repo.stargazers_count / 3 + repo.open_issues_count / 5;
```
spier:
```suggestion
    iScore += repo.forks_count * 5;
    iScore += (repo.subscribers_count ? repo.subscribers_count : 0);
    iScore += repo.stargazers_count / 3;
    iScore += repo.open_issues_count / 5;
```

Now that I am readying this with fresh eyes, I am wondering if it would be easier to understood if we break up this calculation into multiple steps? This way it would be more obvious how each of the properties is weighted. What do you think?

spier:
Thanks for the contribution @3cpt!

Very cool that you are reviewing the pseudo code here.
I am not even sure if it was meant to represent JS but you are probably right. 

You mentioned elsewhere that you are planning to add a score calculation function to one of the [reference implementations of the crawlers](https://github.com/SAP/project-portal-for-innersource/blob/main/docs/CRAWLING.md#reference-implementations), right?

Great catch with the watchers/stargazers/subscribers topic. I need to learn more about this. I am a bit confused after reading the article that you referenced above. Maybe I just got to read it again. :)

I left one [inline comment](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/341/files#r688900183) for you, to see if we could improve readability of the score calculation even further.

Not sure how we could go about validating the pseudo code here.

@Michadelic you will find this one interesting. Any feedback and ideas what we could use to validation the algorithm?

3cpt:
> You mentioned elsewhere that you are planning to add a score calculation function to one of the [reference implementations of the crawlers](https://github.com/SAP/project-portal-for-innersource/blob/main/docs/CRAWLING.md#reference-implementations), right?

Yep. That's why I start analyzing the algorithm, I'm implementing a Python version to integrate in the [zkoppert/innersource-crawler](https://github.com/zkoppert/innersource-crawler/issues/17). Should we add a section in the pattern to implementations of the algorithm?

> Great catch with the watchers/stargazers/subscribers topic. I need to learn more about this. I am a bit confused after reading the article that you referenced above. Maybe I just got to read it again. :)

As far as I understood, they changed watchers to stars and that created a small problem  ([More](https://developer.github.com/changes/2012-09-05-watcher-api/) and [Other](https://github.community/t/bug-watchers-always-same-as-stargazers-count-in-api-user-starred/14224))

> Not sure how we could go about validating the pseudo code here.

Agree. What about use the mock data from the [planet's repo.json](https://github.com/SAP/project-portal-for-innersource/blob/main/repos.json) and generate the score instead of mock it too?

spier:
@3cpt just confirming that I didn't forget about your PR. Would just like to wait for @Michadelic for feedback.
(it might still be holiday season in some places so let's give him some time :)).

3cpt:
@spier No worries. I replicated the same code in Python and C# to validate the consistency of the numbers.
So if someone uses the algorithm to gather the score, please, share it! Real data is always good in this case.

Michadelic:
Hi @3cpt @spier, indeed there is holiday season over here so please excuse my late response. 
Thank you very much for your great findings, will take a look in detail now and provide feedback in the code directly.

I took this code directly from our running crawler on GitHub enterprise, so it is a working JavaScript implemention, but for sure it can be further tweaked and adjusted. If you spot more issues just let me know.

Michael

patterns/2-structured/repository-activity-score.md
```diff
@@ -58,15 +58,18 @@ function calculateScore(repo) {
     // initial score is 50 to give active repos with low GitHub KPIs (forks, watchers, stars) a better starting point
     let iScore = 50;
     // weighting: forks and watches count most, then stars, add some little score for open issues, too
-    iScore += repo["forks_count"] * 5 + repo["watchers_count"] + repo["stargazers_count"] / 3 + repo["open_issues_count"] / 5;
+    iScore += repo.forks_count * 5;
+    iScore += (repo.subscribers_count ? repo.subscribers_count : 0);
+    iScore += repo.stargazers_count / 3;
+    iScore += repo.open_issues_count / 5;
```
Michadelic, line 64: great, thanks for this!

patterns/2-structured/repository-activity-score.md
```diff
@@ -58,15 +58,18 @@ function calculateScore(repo) {
...
     let iDaysSinceLastUpdate = (new Date().getTime() - new Date(repo.updated_at).getTime()) / 1000 / 86400;
     // updated in last 3 months: adds a bonus multiplier between 0..1 to overall score (1 = updated today, 0 = updated more than 100 days ago)
     iScore = iScore * (1 + (100 - Math.min(iDaysSinceLastUpdate, 100)) / 100);
     // evaluate participation stats for the previous  3 months
     repo._InnerSourceMetadata = repo._InnerSourceMetadata || {};
     if (repo._InnerSourceMetadata.participation) {
         // average commits: adds a bonus multiplier between 0..1 to overall score (1 = >10 commits per week, 0 = less than 3 commits per week)
-        let iAverageCommitsPerWeek = repo._InnerSourceMetadata.participation.slice(repo._InnerSourceMetadata.participation - 13).reduce((a, b) => a + b) / 13;
+        let iAverageCommitsPerWeek = repo._InnerSourceMetadata.participation.slice(13).reduce((a, b) => a + b) / 13;
```
Michadelic: i think the order in which the GitHub API returns it is reversed, therefore i took the end - 13 instead of the first 13 entries. I never left a comment on this behavior, however. So its kind of hard to understand this line, agreed.

In the documentation is stated that the number of commits for the previous 52 weeks is returned, but we only consider the last 13 weeks (~= 3 Months), meaning the last 13 entries of the array returned by the API.

This should remain as is in my opinion. Maybe we can add a comment about the behavior:
"The array order is oldest week (index 0) to most recent week."

Michadelic (changes requested):
Great simplifications, just the slice change should be reverted (see my comment)

patterns/2-structured/repository-activity-score.md
3cpt: [The array order is oldest week (index 0) to most recent week.](https://docs.github.com/en/rest/reference/repos#get-the-weekly-commit-count)

Understood, my fault. I will revert that change.

patterns/2-structured/repository-activity-score.md
3cpt: It was @spier work 👏

spier:
@Michadelic would you mind reviewing/approving this PR once it is in a state that is ready to be merged?
I don't feel knowledgeable enough myself about the activity score to do this myself. Cheers.

patterns/2-structured/repository-activity-score.md
3cpt: Updated. And I also added the `.length` because at least during my tests without that it gets all the data, and doesn't slice the array.

3cpt:
During my last check I verified that the `+ 1` in the L64 and L70 (in the previous version) lead the result of the multiplier to `> 1`, when the rule in the comment says that should be between `0...1`.
So, if, in the worst case scenario we get 0, `0 / x = 0`, if it is `x / x = 1`.

Please, your feedback is welcome, and I could be doing my math wrongly.

patterns/2-structured/repository-activity-score.md
```diff
@@ -58,17 +58,23 @@ function calculateScore(repo) {
...
+    iScore = iScore * ((100 - Math.min(iDaysSinceLastUpdate, 100)) / 100);
+
     // evaluate participation stats for the previous  3 months
     repo._InnerSourceMetadata = repo._InnerSourceMetadata || {};
     if (repo._InnerSourceMetadata.participation) {
         // average commits: adds a bonus multiplier between 0..1 to overall score (1 = >10 commits per week, 0 = less than 3 commits per week)
-        let iAverageCommitsPerWeek = repo._InnerSourceMetadata.participation.slice(repo._InnerSourceMetadata.participation - 13).reduce((a, b) => a + b) / 13;
-        iScore = iScore * (1 + (Math.min(Math.max(iAverageCommitsPerWeek - 3, 0), 7)) / 7);
+        let iAverageCommitsPerWeek = repo._InnerSourceMetadata.participation.slice(repo._InnerSourceMetadata.participation.length - 13).reduce((a, b) => a + b) / 13;
+        iScore = iScore * ((Math.min(Math.max(iAverageCommitsPerWeek - 3, 0), 7)) / 7);
```
Michadelic: Hi , @3cpt! I agree the comment with 0..1 is a bit irritating, but on top of whats already there, we want to add a multiplier that scales the result based on the frequency of the current commits. If we remove the  "1+ multiplier" then the whole result is scaled down instead of adding something on top (e.g.  100 * 0.5 = 50 whereas 500 * (1 + 0.5) = 750). Therefore, let's keep the implementation as is for this line and rather work on the comment if it is irritating. Hope this explanation makes sense!

Michadelic (changes requested):
please keep the "1+" as explained in the comment, other changes for readability are appreciated :+1:

patterns/2-structured/repository-activity-score.md
3cpt: Reverted. But, in my opinion, should be created an issue to clarify the comments.

spier:
@Michadelic would you mind reviewing the latest status of this PR again, and then approving the PR when everything looks good?

To me it looks like @3cpt implemented the change that you asked for (thanks for that!) for but I am not 100% sure so I figured I rather ask you for confirmation.

patterns/2-structured/repository-activity-score.md
spier: @3cpt do you have a suggestion for how the comment would be easier to understand?

You know best what you are looking for here, so by all means, feel free to raise a new ticket to clarify the comment.

3cpt:
> @Michadelic would you mind reviewing the latest status of this PR again, and then approving the PR when everything looks good?
> 
> To me it looks like @3cpt implemented the change that you asked for (thanks for that!) for but I am not 100% sure so I figured I rather ask you for confirmation.

By removing the 0…1. Because the multiplier will be higher.

patterns/2-structured/repository-activity-score.md
3cpt: By removing the 0…1. Because the multiplier will be higher. But, yes, I believe that could be discussed in a future issue

Michadelic (approved):
Thank you for the improvements, looks good now!

Michadelic:
@spier PR can be merged now, looking good

spier:
Thank you for this contribution @3cpt !!! Much appreciated fix.

Also we are looking forward to see your reference implementation.
Once that is live, then we could also link to that from this pattern.

Issue #162, opened 2020-05-06
Auto-Generate the Mindmap on .md-source changes


lenucksi:
#156 is great stuff, we just need to remember to add a GitHub Action to regenerate the SVG / HTML / rendered products once the .md changes and make a checklist item for PRs to update the .md once new things come in to keep everything fresh. 

_Originally posted by @lenucksi in https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/156#issuecomment-624834371_

spier:
When you say mindmap, are you referring to `IS.md` and `IS.html`?

What I am missing from `IS.md` as well is an explanation of what it is. e.g. if any reader of the IS Patterns finds this file, how would they know what it is?

lenucksi:
> When you say mindmap, are you referring to `IS.md` and `IS.html`?

With `mindmap` I'd refer to the rendered artifact(s), i.e. the SVG or the SVG embedded in a HTML file.
`mindmap source` is the underlying MD file.

Sorry for having an issue that is still a bit raw - it's generated from a PR comment to be dealt with later.
> 
> What I am missing from `IS.md` as well is an explanation of what it is. e.g. if any reader of the IS Patterns finds this file, how would they know what it is?

I would hope that we have the mindmap embedded an a larger production (e.g. the pattern catalogue or book at some point) that has additional details explaining it.

spier:
FYI documentation and files related to the mindmap are now in:
https://github.com/InnerSourceCommons/InnerSourcePatterns/tree/master/pattern-categorization

spier:
I asked a question on the [markmap repo](https://github.com/gera2ld/markmap/issues/66) related to this. 

While we could generate the input file (markdown) automatically, markmap doesn't have an option to turn that into a static file that we can embed in our book at https://patterns.innersourcecommons.org.

spier:
This particular issue has been implemented in #348

PR #348, opened 2021-08-30
Generate mind map automatically (via GHA)

spier:
This is part of the solution for #162. 

Assuming that a fully automated solution would consist of these steps:
1) consolidate the categories from all `patterns/*.md` files into `innersource-program-mind-map.md` (this is different from how this is working today)
2) generating the mindmap HTML (`innersource-program-mind-map.html`)
3) generating a screenshot from that HTML, that can be used in our book

This PR solves 2 + 3.

## Tasks

- [x] update README
- [x] decide what to do with `book/innersource-program-mind-map.PNG` => we are now just keeping this file in the folder `/pattern-categorization`, and link to it from `/book`
- [x] limit GHA execution to cases where the source files for markmap have been changed
- [x] make the GHA execution faster (possibly with caching dependencies via [actions/setup-node](https://github.com/actions/setup-node)) - tried this for a bit but without major time savings

## Update

I noticed that solution step (1) above wasn't necessarily intended as part of #162.
Hence I will leave it out of this PR, and instead file a new issue that describes the idea behind (1) in more detail.


spier:
@fwan2000 you may find this interesting. I found a way to use a GitHub action to autogenerate the mindmap on any change to the source file. See [here](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/348/files?file-filters%5B%5D=.html&file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.png&file-filters%5B%5D=.yml&file-filters%5B%5D=dotfile&hide-deleted-files=true#diff-b049987d441c67db672493cdfab3b4268c8874e100405d6623836b616cdb4fc5).

lenucksi:
Cool stuff! 🎉

PR #358, opened 2021-09-20
Formatting fixes for the "Introducing Metrics in InnerSource" pattern

spier:
- Change labels for **Problem**/**Context** section (I am pretty sure they were the wrong way around)
- Change **Forces**/**Context** to bullet list
- Break out 'Metrics Candidates' into a dedicated subsection below the **Solution** section
- Using InnerSource instead of IS


spier:
As these are only minor formatting fixes, I am planning to leave this open until Wed/Thu this week, and then merge it.
Unless I get an approval review before that of course :)

spier:
Actually I changed my mind on this.
As this only contains formatting fixes and non-controversial spelling fixes, I will merge this right away, to prevent any potential merge conflicts with other improvements to this pattern.

FYI @dicortazar

PR #360, opened 2021-09-23
Renaming the keys for the 'Support and Maintenance' group

spier:
I was reading the Maturity Model, and noticed that the keys of two groups were colliding with each other.
Both groups **Sharing Policies** and **Support and Maintenance** were using keys prefixed with `SP-`.

I am assuming here that the keys are supposed to be unique within the file. If that is not the case, please disregard this PR :)

Changes made in this PR:
* renamed keys for the **Support and Maintenance** group to `SM-` stead.

Command used to extra only the rows with the rules/keys from the file (mostly a mental note for myself):
`ack "\* .{2}\-\d" maturity-model.md | sort`

### DIFF
--- a/patterns/2-structured/maturity-model.md
+++ b/patterns/2-structured/maturity-model.md
@@ -175,10 +175,10 @@ InnerSource projects need a means for self assessment. Metrics can be one aspect
 
 Not only should feature development be owned by InnerSource teams - support and maintenance is also part of the teams core tasks.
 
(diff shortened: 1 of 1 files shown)


dicortazar:
Thanks a lot @spier. This works for me and thanks for hunting the naming issue. @MaineC any further comment here?

dicortazar (approved):
This change makes sense as we do not use the same acronym in two places.

MaineC:
No further comment, looks good to me.

PR #380, opened 2022-01-08
Switching to markdown syntax of pictures in all Structured Patterns

spier:
Switching to markdown syntax of pictures in all Structured Patterns, as otherwise the alias doesn't get displayed correctly as a caption below the picture in gitbook.


spier:
Minor change. Merging without review.

spier:
Turns out that as a side-effect of this change, the pictures are now shown in the maximum width available on the page.

e.g. as seen with the picture in https://patterns.innersourcecommons.org/p/contracted-contributor

image size example 2022-01-08

PR #383, opened 2022-01-08
Fix duplicate header on ToC page

spier:
The table of contents (ToC) page in the book shows the same header twice.

This is only happening to the ToC but not for other default files that gitbook uses (like `introduction.md`).
I am moving the headline in the ToC file to the very top, to see if it gets handled correct by git book then (i.e. I expect to only see the title "Table of Contents" once).

Current (buggy) behavior:

Screenshot 2022-01-08 at 19 18 25

### DIFF
--- a/book/toc_template.md
+++ b/book/toc_template.md
@@ -1,14 +1,14 @@
+# Table of Contents
+
 <!--
 Do not edit toc.md directly!!!
 Instead edit toc_template.md
 -->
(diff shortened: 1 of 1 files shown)


spier:
Yep, that fixed the issue.

Now bitbook correctly shows this:

Screenshot 2022-01-08 at 19 26 30

PR #387, opened 2022-02-08
Fix linter issue "Emphasis style should be consistent"

spier:
Fix linter issue "Emphasis style should be consistent".

**Aside:**
This is a new issue reported by markdownlint in
https://github.com/InnerSourceCommons/InnerSourcePatterns/runs/5112470503

We did not get this linter error before, and this particular pattern did not change recently.
So not sure why markdownlint is complaining now :)


spier:
Fixing a minor markdown validation issue in **main**.
Merging without review.

Issue #386, opened 2022-02-08
Missing patterns related to Common Requirements pattern


jsmanrique:
Reading the "[Common Requirements](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/patterns/2-structured/common-requirements.md)" pattern I've found there are a couple of patterns mentioned as "pattern to be written", but I haven't found any additional information about them.

The missing patterns seem to be related to the initial and resulting context of the Common Requirements pattern, so I find them quite important to fully understand the pattern itself:

From the pattern Context section:
```text
This pattern applies if there is a Strong Code Owner [pattern to be written] or if there is weak code ownership, 
or no Benevolent Sponsor [pattern to be written].
```

From the Resulting Context section:
```text
A related pattern [pattern to be written] is a circular story-writing exercise reported at one company 
employing InnerSource
```

Wouldn't it be nice to have those patterns drafted?

Where could we look for related information?

spier:
Great catch @jsmanrique!

Based on the [requirements](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/meta/contributor-handbook.md#requirements-level-2---structured) that we are applying on patterns today, this pattern should be in maturity level "Structured". If we were to write this pattern today, we would keep it in "Initial" i.e. we would not publish it to the book until the placeholders are removed.

However I believe that this pattern was written and published at a time before we had such written requirements and review guidelines in place. So in a way it is an artifact from the early days of the InnerSourcePatterns repo :)

I will try to find traces of the patterns that are linked to from the [Common Requirements](https://patterns.innersourcecommons.org/p/common-requirements) pattern. There is an old Google Group that might have some traces but I don't know if it is publicly available.

And I see that you already raised this question in Slack as well, so maybe we can find somebody who has further context from the time at which the pattern was written.

spier:
I found the PR in which this pattern was first created:
https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/11

That PR also contains a mention of "PLoP 2017", which I assume might have been an in-person event of the group at which the pattern was written.

Still looking for more ...

spier:
When we re-orgnized the files in our repo the pattern file `common-requirements.md` was moved (without doing much change to the content itself).

[...]

PR #388, opened 2022-02-10
Removing placeholders from Common Requirements pattern

spier:
This PR addresses some issues with the Common Requirements pattern that were raised in #386. 

### Key changes:

- Adding an anonymized Known Instance (as that is a content requirement for Structured patterns)
- Split Context into individual bullets
- Context: Replacing placeholders with short definitions and highlighting key terms
- Resulting Context: Replacing placeholder 
- Using enumeration to emphasize that that the solution consists of two aspects

### Ideas for future improvements:

There are some other quality issues with this pattern that are not addressed by this PR.
I left those out on purpose, to keep the review a bit lighter.

These things could be improved in future PRs:

- Improve the **Patlet** - escpecially describe how the solution is improving upon the problem
- Break up **Forces** into bullets - so that readers can check more easily if they have these forces as well
- Decide what to do with the **Resulting Context**: It currently focuses mostly on other related challenges that may need to be solved after this pattern is in place. However it lacks explanation of how this pattern has helped to address the problem.
- improving the [visual](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/assets/img/CommonReqtsv2.jpg) e.g. by creating a new version digitally (so no scan), and explaining in the patttern what that visual illustrates

### DIFF
--- a/patterns/2-structured/common-requirements.md
+++ b/patterns/2-structured/common-requirements.md
@@ -17,9 +17,9 @@ The common code in the shared repository isn't meeting the needs of all the proj
 * The common code is a small percentage of the overall deliverable from any of the projects.
 * Each project has its own delivery schedule, set of deliverables and customers.
 * This pattern applies in either of these situations:
-    * there is a Strong Code Owner i.e. all changes to the shared repository have to be approved by the repo owner
-    * there is weak code ownership i.e. no one really owns the code
-    * there is no Benevolent Sponsor i.e. no organization or executive is providing resources to organize the inner sourcing project, essentially taking de facto ownership of this common code.
+    * there is a **Strong Code Owner** i.e. all changes to the shared repository have to be approved by the repo owner
+    * there is **weak code ownership** i.e. no one really owns the code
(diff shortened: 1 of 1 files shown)


spier:
@jsmanrique would you mind reviewing this PR, to see if it helps with the issues you raised in #386 ?

NewMexicoKid (approved):
Proposed changes look okay to me. Do note that there are many different styles of writing patterns; back in the old days, expressing the context in paragraph form rather than in separate bullets was thought to be okay. That being said, I understand the value of breaking it down.

spier:
Thanks for the review!

Absolutely, I also think paragraphs are ok in the Context. 
Personally I prefer one paragraph for each precondition in that case, as it makes it easier for the reader to use the Context as a checklist.

Even a mix might good in situations
e.g. some paragraphs in the beginning of the Context (kind of like a general intro) and then a bullet list of preconditions. 

I will see if I can get one more review and then merge this PR.

Thanks again!

jsmanrique:
ILGTM. Thank you very much for reviewing this and working on the PR!

spier:
Awesome. The improved pattern is now live at:
https://patterns.innersourcecommons.org/p/common-requirements

PR #397, opened 2022-03-01
Add GitHub to known instances

zkoppert:


spier (approved):
Thanks for adding a new Known Instance!

spier:
@zkoppert this is already merged and live at https://patterns.innersourcecommons.org/p/30-day-warranty.
Thank you! 🚀 

Is there anything extra that you want to share about how this pattern is applied at GitHub? e.g.
- why the 6 week?
- how do people talk about it internally? do they say "warranty" or does it go by a different name?
- are there any activities that the teams do at the point when the warranty runs out? (i.e. when the 6 weeks are over)
- where do you communicate the "rules" of the warranty? is it in the Contribution Guidelines or elsewhere? 
- anything else about those rules that isn't covered in this pattern already?

If any of these are triggering some thoughts that you find are relevant to share, please do.
We would happily review another PR :)

PR #399, opened 2022-03-01
Add Microsoft's use of 30 day warranty pattern

arnom-ms:
Add Microsof's use of the 30 day warranty pattern

### DIFF
--- a/patterns/2-structured/30-day-warranty.md
+++ b/patterns/2-structured/30-day-warranty.md
@@ -48,6 +48,7 @@ In addition it helps to provide clear [contribution guidelines](./project-setup/
 
 - This was tried and proven successful at PayPal.
 - GitHub internally uses this pattern with a modified warranty timeline of 6 weeks.
+- Microsoft recommends this pattern as a principle - teams set their own specific time target matching their needs and confidence.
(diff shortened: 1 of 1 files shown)


robtuley:
Thanks @arnom-ms great to add to the known instances of this pattern -- your update is now live at https://patterns.innersourcecommons.org/p/30-day-warranty#known-instances 

Not sure if you saw Sebastian's comment in a recent similar PR, same applies...

> Is there anything extra that you want to share about how this pattern is applied at Microsoft? e.g.
> what warranty durations do teams typically ask for? 
> how do people talk about it internally? do they say "warranty" or does it go by a different name?
> are there any activities that the teams do at the point when the warranty runs out? (i.e. when the 6 weeks are over)
> where do you communicate the "rules" of the warranty? is it in the Contribution Guidelines or elsewhere?
> anything else about those rules that isn't covered in this pattern already?
> If any of these are triggering some thoughts that you find are relevant to share, please do.
> We would happily review another PR :)

spier:
Thank you @arnom-ms for the contribution. 
And great job @robtuley for handling this PR so quickly.

PR #404, opened 2022-04-13
Fix GitHub links in Repository Activity Score pattern

spier:
GitHub moved some of the REST API documentation.
This fixes the links in the Repository Activity Score pattern.

### DIFF
--- a/patterns/2-structured/repository-activity-score.md
+++ b/patterns/2-structured/repository-activity-score.md
@@ -46,7 +46,7 @@ The repository activity score is a numeric value that represents the (GitHub) ac
 In addition, it considers activity parameters like last update and creation date of the repo to give young projects with a lot of traction a boost.
 Projects with contributing guidelines, active participation stats, and issues (public backlog) receive a higher ranking as well.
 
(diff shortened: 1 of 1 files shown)


spier:
The link checker still returns an error.

```
ERROR	https://docs.github.com/en/rest/reference/metrics#statistics
	Forbidden (HTTP error 403)
ERROR	https://docs.github.com/en/rest/reference/search#search-repositories
	Forbidden (HTTP error 403)
```

Not sure what's up there.

spier:
@zkoppert maybe you can help me with a broken link problem of my own, given that these are github.com links? 😈 

In this repo we use an automatic link checker, which spotted that the old links were broken.
However when fixing them to e.g. `https://docs.github.com/en/rest/reference/metrics#statistics`, the link checker reports a **403** status for these.

Do you know if maybe it isn't allowed (by GitHub) to do automatic link checks on pages below `docs.github.com`?

zkoppert:
@spier I don't see any reason why the link checker should have failed. If you have permissions, Can you rerun the action to see if it was a temporary error. We have had some hosting interruptions recently and I want to rule that out here if possible.

spier:
Thanks @zkoppert , that was a great lead!
I re-ran the link checker GHA and now all checks are passing.

PR #405, opened 2022-04-13
Recommendations for using InnerSource as a verb in a sentence

zemunk:
The proposed changes are based on the following Slack thread: https://innersourcecommons.slack.com/archives/C04PXKRN4/p1649258602339809

The changes include:
- example sentences when InnerSource is used as a verb
- proposed correction of camel-case to 'Pascal case' as it is a more frequently used term than 'upper camel case'

**Note:**
In the long term it would be nice to spell InnerSource more like open source and maybe allow using it as a verb too (just because it rolls off the tongue).

### DIFF
--- a/meta/innersource-spelling.md
+++ b/meta/innersource-spelling.md
@@ -10,7 +10,7 @@ We encourage you to use the word as a proper noun (like “We published the repo
 
 We favor the spelling **InnerSource**, for the following reasons:
 
-1. That’s the way Tim O’Reilly [spelled it in 2000][opengl_1200] (he was into Perl and Pascal case was a thing). Also see the [Foreword of the Adopting InnerSource book][foreword_AdoptingInnerSource] for more of Tim O’Reilly's own words on the topic.
+1. That’s the way Tim O’Reilly [spelled it in 2000][opengl_1200] (he was into Perl and camel-case was a thing). Also see the [Foreword of the Adopting InnerSource book][foreword_AdoptingInnerSource] for more of Tim O’Reilly's own words on the topic.
 2. If you set up side-by-side Google searches for “Inner Source” and “InnerSource” you will find that you get more hits on the former term, by only 1% of them have anything to do with what we call InnerSource. The latter term will be 100% cogent to your inquiry.
 3. The OSI was [denied trademark on the term “open source”][no-open-source-trademark] because it was made up of two common and unrelated terms. “InnerSource” on the other hand is a new word.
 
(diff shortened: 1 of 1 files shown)


spier:
Thanks for the suggestions @zemunk. Will try to take a look shortly.

meta/innersource-spelling.md
```diff
@@ -2,15 +2,15 @@
...
+We encourage you to use the word as a proper noun (like “We published the repository as InnerSource”) wherever possible. We discourage the use of verb forms of the word (like “We’re InnerSourcing that repo”).
 
 ## Full story - Why we care about the spelling
 
 **InnerSource** is an artificial term that does necessarily have a spelling that would naturally be correct. Therefore you will find various spellings when you search for the term online.
 
 We favor the spelling **InnerSource**, for the following reasons:
 
-1. That’s the way Tim O’Reilly [spelled it in 2000][opengl_1200] (he was into Perl and camel-case was a thing). Also see the [Foreword of the Adopting InnerSource book][foreword_AdoptingInnerSource] for more of Tim O’Reilly's own words on the topic.
+1. That’s the way Tim O’Reilly [spelled it in 2000][opengl_1200] (he was into Perl and Pascal case was a thing). Also see the [Foreword of the Adopting InnerSource book][foreword_AdoptingInnerSource] for more of Tim O’Reilly's own words on the topic.
```
spier: Just curious: Why the preference for "Pascal case" over "camel case"?

meta/innersource-spelling.md
```diff
@@ -24,15 +24,15 @@ For new patterns we stick to the spelling **InnerSource**. Old patterns will be
...
 
 We recommend not to turn the term into a verb, but rather using it as a noun or adjective.
 
 Recommended use:
 
-“We use InnerSource in that repo”
+“We published the repository as InnerSource.”
 or
-“We use InnerSource methods in that repo”
+“We published the repository under an InnerSource license.”
```
spier:
```suggestion
“We use InnerSource methods in that repo.”
```

I would prefer to keep the original example here, as using InnerSources isn't that common (from what I know), so I don't want to suggest that they are.

However if you want to add any other example besides licenses, please do. Doesn't hurt to have 3-4 or even more examples of recommended use.

meta/innersource-spelling.md
zemunk: I knew camel-case to start with lower case letters e.g. `someWords` and those were the first results in Google for me. upon further research it seems to be synonymous to Pascal case so I will revert these changes 🙂

meta/innersource-spelling.md
zemunk: I kept the example without the license and removed the other one.

spier (approved):
Thank you for these improvements!

At least the InnerSource Commons has pretty specific reasons to use the spelling `InnerSource` but we could probably do a better job explaining them. As an aside, there is already some work underway for that.

As for using it in verb-form, I agree. 
It can feel clunky at times to just use it as a noun.
Note quite sure how to do that best though.

PR #407, opened 2022-04-25
Adding sketch to RFC pattern

spier:
Sketch is from this [blog post](https://blog.pragmaticengineer.com/scaling-engineering-teams-via-writing-things-down-rfcs/) by Gergely Orosz.

I confirmed here that it is ok to use it:
https://twitter.com/GergelyOrosz/status/1518296942819909634


spier:
I am adding an image from a source that we have already referenced under Known Instances.
Therefore I am merging without review.

PR #408, opened 2022-04-29
Switch to canonical wording for InnerSource

MaineC:
A long time ago we decided to call things "InnerSource" - written without space and in camel case. This pattern seems to pre-date this decision, so changing.

One note to self: There is an empty reference to "contribution negotiation" - I believe some of the teams at Europace already found out that in addition to contribution guidelines and warranty the entire contribution process starts before the submission of the code change. IIRC something like that was also part of our contributor training that I did with @lenucksi. Likely we already have the knowledge for that pattern in the community.

### DIFF
--- a/patterns/1-initial/share-your-code-to-get-more-done.md
+++ b/patterns/1-initial/share-your-code-to-get-more-done.md
@@ -8,8 +8,8 @@ TBD
 
 ## Problem
 
-* Developers do not realize that they can get more done by taking advantage of Inner Sourcing
-* Developers continue to work in their silo or alone and never take the time invest in moving to Inner Sourcing
+* Developers do not realize that they can get more done by taking advantage of InnerSource
+* Developers continue to work in their silo or alone and never take the time invest in moving to InnerSource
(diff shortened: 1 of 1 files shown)


lenucksi:
> One note to self: There is an empty reference to "contribution negotiation" - I believe some of the teams at Europace already found out that in addition to contribution guidelines and warranty the entire contribution process starts before the submission of the code change. IIRC something like that was also part of our contributor training that I did with @lenucksi. Likely we already have the knowledge for that pattern in the community.

Yes, that should be part of the contribution mechanics section.

spier (approved):
LGTM

spier:
Thanks for this fix @MaineC. 
I did merge this already even though the "Link Checker" is failing, as that is a sporadic issue that the GitHub URLs seem to have. Hope it resolves itself.

For the note about "contribution negotiation":
If you want to create an issue for this with a specific activity that we could do, that would be great.

MaineC:
@spier good suggestion to move that into a separate issue - done that there: https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/410

PR #368, opened 2021-11-18
Pattern idea: Crossing the chasm

MaineC:
On slack a presentation was shared on the adoption cycle for InnerSource: https://innersourcecommons.slack.com/archives/C04PXKRN4/p1636027720045600?thread_ts=1636027623.045400&cid=C04PXKRN4

This pull request is a first draft at providing a pattern for what InnerSource initiatives can do to reach teams and people at different places on the adoption curve.

ping @rrrutledge

PR #414, opened 2022-05-14
Title-case pattern name and add link from README

spier:
Follow-up to #368.

As an aside:
Would be nice to automatically list all patterns in the main `README.md`, so that PRs like this would not be necessary.


spier:
The Link Check errors are know issues.

As this is an obvious change, I am merging without review.

PR #415, opened 2022-05-15
Board report 2022 05

spier:


spier:
The Board meeting where this report will be discussed is next Tuesday.
Hence merging this.

PR #422, opened 2022-06-07
Adding the initial Japanese translation

yuhattor:
I translated all the structured patterns as of 06/07/2022 into Japanese. This will be helpful to understand InnerSource practices for Japanese. I am considering the possibility of managing them under InnerSourceCommons so that Japanese users can easily find them.

The sample of the deliverables at this time can be found [here]( https://george-15.gitbook.io/innersource-patterns-in-japanese/p-1/japanese).

This PR includes:

* [x] Translated all the structured patterns into Japanese. The translation can be found under `/translation/japanese` 
* [x] Proofread all the document
* [x] Linted markdown files
* [x] Edited the ToC generation script and templates

Things to consider:

* [x] How to provide the Japanese translation
  Posting all translations on [InnerSource Patterns](https://patterns.innersourcecommons.org/) page is good for Japanese. However, most users will find it unpleasant to see Japanese in their search results on GitBook. Also, it is not good that there is noise in the table of contents.

Alternative ways to provide Japanese translation:

* Create a Japanese page on a different domain such as jp.patterns.innersourcecommons.org or patterns-jp.innersourcecommons.org. However, this method requires double management of GitBooks.
* Maybe we could use GitBook's feature to exclude from searches within GitBook and foldable ToCs. Did a little research and it seemed difficult. Need to do more research.

This translation could be a private project for me, but I think it would be better managed somehow under the InnerSource Commons in order to promote the InnerSource Community in Japan.

Thanks!

PR #425, opened 2022-07-14
updated toc japanese translation

yuhattor:
@spier 
#422 does not have this ToC, so I've made a PR for the ToC translation for this branch.


spier (approved):
LGTM

Thanks a lot @yuhattor!

PR #426, opened 2022-07-20
gitbook integration for Japanese Translation v2

spier:
This PR takes the fantastic Japanese translation work in #422, and adds the pieces to integrate English and Japanese side by side in gitbook.

PR #427, opened 2022-07-21
Fixing links in English book (after moving files)

spier:
As part of https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/426 we are moving the files for the EN gitbook to a new folder `book/en`. Therefore some relative links in the markdown files are likely broken, which we are fixing in this PR.

Update:
Also updated the versions of some GHAs used in `.github/workflows/markdownlint.yml`.


spier:
I am merging this back to `book-jp` without review, as that will be the branch that then gets merged to `main` eventually.
However I will keep this `book-en` branch around, as this is where we render the EN book from on Staging (at least for now).

PR #431, opened 2022-07-28
Improve docs of book/translation/contribution

spier:
Hi @yuhattor. The automatic PR assignment is working (as you can see in the Reviewers on the right-hand side).

I made some improvements to the explanation of
a) our book publishing process (removing info about the `book-staging` branch because we don't actually use that any more)
b) translation process - simplified some of the titles by removing "how to ..."
c) the contribution guidelines - pointing potential contributors to our translations to advertise that as one possible way to contribute

As you have recently gone through both the translation and contribution process, I would be very grateful if you could review this PR :)


yuhattor (approved):
Thank you @spier for updating the README files:)

yuhattor:
Sorry, a little late to respond!
I've merged!

spier:
Thanks for the review!

And no worries, it is completely ok if reviews take days, some even weeks, given that we are all donating time here to make this work.

PR #435, opened 2022-08-09
Make all gitbook URLs the same in JP and EN book

spier:
Most gitbook URLs are the same between EN and JP already.

However not the URLs for the pages in the Appendix section: 
- https://patterns.innersourcecommons.org/v/jp/fu-lu/pattern-template
- https://patterns.innersourcecommons.org/v/jp/fu-lu/sono/readme-template
- https://patterns.innersourcecommons.org/v/jp/fu-lu/sono/contributing-template

### Why is that?

Normally the URLs are generated from the filename (which is the same for JP and EN content).
However the sections Appendix and Appendix/Extras or no files but rather groups of pages within the menu of the gitbook.
Therefore gitbook is likely auto-generating the URL based on the name of these groups.
As the group names are different for JP and EN, we end up with different URLs.

### Do we care?

It would be nice to keep the URLs the same between all books, so that:
- users can switch easily between the languages and stay on the same page, just in a different language
- content creators can more easily go back and forth between languages to check if the content looks consistent across different languages

### How to fix it?

I hope that we can fix it by setting explicit IDs in the table of contents.
e.g. `<a id="appendix">` or similar.

I am trying this out in this PR.


spier:
The fix works as expected for "Appendix".
But not for "Appendix/Extras".

spier:
[4089dcc](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/435/commits/4089dcc46f4bcdc7986a216a76ee1f520c85b3f7) did not work as desired.

Rather than creating a subgroup in the menu (see 1) it creates an entirely new group and also different URLs (see 2).
So that does not help us.

1:
Screenshot 2022-08-09 at 11 04 02

2:
Screenshot 2022-08-09 at 11 04 12

spier:
I am sending this question to gitbook support now:

> I am working on this gitbook:
> https://innersourcecommons.gitbook.io/innersource-patterns-staging/v/jp-test-urls/appendix/sono/readme-template
> 
> In the URL above, is it possible to replace the `sono` with something else? I want it to be `extras` instead as that is what we use in the English version of the same book. That way I could keep the URLs the same between all books.
> 
> It looks like gitbook is auto-generating the URL based on the name of the groups in my TOC/menu. And I know that I can set explicit IDs to modify this as in <a id="appendix"></a>.
> 
> However in the example above the pages under `sono` are not in their own group but rather in a sublist.
> 
> Code snippet:
> 
> ```
> ## 付録<a id="appendix"></a>
> 
> * [パターンテンプレート](../../meta/pattern-template.md)
> * その他<a id="extras"></a>
>   * [README テンプレート](../../translation/japanese/templates/README-template.md)
>   * [CONTRIBUTING テンプレート](../../translation/japanese/templates/CONTRIBUTING-template.md)
> ```
> 
> As you see I already tried to set an ID for that sub-list as well but gitbook does not pick that up.
>
> Any other ideas of what I could do to keep the URLs the same between all translations of the book?

spier:
The gitbook Support team recommend to force a URL change manually via the gitbook UI.
I tried that, which resulted in [9c0d86b](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/435/commits/9c0d86b1cd87485113b1d70e415717291effb30a).

Not sure if we can use this approach, as it messes too much with our file structure.
Have to go through it in more detail to understand it.

spier:
I feel like I am getting too far in reverse-engineering how gitbook constructs the URLs based on path + filename + other factors. Maintaining such customizations would be too cumbersome.

Hence I am getting this PR into a state where we only fix the URL for pattern-template but leave the other ones as they are.

spier:
I am keeping this branch around for now.

Issue #434, opened 2022-08-08
Improve the Japanese translation of InnerSource Patterns


yuhattor:
Japanese Translation of InnerSource Patterns is in the early phases. InnerSource Commons Japan has adopted textlint, including unifying Japanese notation and improving readability, and discussions are being held by volunteer members. When the Japanese translation has been completed to some extent, we will improve the Japanese of InnerSource Patterns by referring to the TRANSLATION-POLICY and textlint in the following repository.
https://github.com/InnerSourceCommons/japanese-contents

spier:
I am not 100% clear what improvements this issue is describing. 

Trying to summarize what I understood from this issue:

* link from [InnerSourcePatterns/../translation/README.md](https://github.com/InnerSourceCommons/InnerSourcePatterns/tree/main/translation) to [japanese-contents/TRANSLATION-POLICY.md](https://github.com/InnerSourceCommons/japanese-contents/blob/main/TRANSLATION-POLICY.md)
* apply [japanese-contents/../textlint](https://github.com/InnerSourceCommons/japanese-contents/tree/main/textlint) in the InnerSourcePatterns repo as well? (I don't know textlint. is it similar to markdownlint? would it make sense to adopt it for the whole InnerSourcePatterns, or just for the japanese translation?)

If you could help clarify this, that would be great.

yuhattor:
We are currently working on standardizing the Japanese notation. And [textlint](https://github.com/textlint/textlint) is lint tool for natural languages. For instance, this tool is helpful when you don't want the words "center" and "center" mixed together in one document.

This issue was created as a task to improve the Japanese translation based on the rules created by the Japanese community, so it is not intended to go so far as to put a specific lint into the CI/CD pipeline. For the time being, I plan to run textlint on the cli. I think textlint is a very good tool, maybe we could implement it in the InnerSource Patterns pipeline someday

spier:
Sounds interesting. I would like to see textlint in action :) 

If I don't get to try it out myself, maybe you could do a little demo at some point?

yuhattor:
For the record, I will do work on the [jp-translation-improvement](https://github.com/InnerSourceCommons/InnerSourcePatterns/tree/jp-translation-improvement) branch. The branch name like book-jp-improvement might be suitable for management, but it looks that `.github/workflows/book.yml` triggers the GH Actions and update .gitbook.yaml because the branch name starts with "book-jp" 👍

spier:
[...]

PR #440, opened 2022-08-21
Japanese translation improvement

yuhattor:
- Fixed and Unified some expressions on Trusted Committer.
- Fixed markdown 
#434


yuhattor:
@spier Review is required, Just a quick review please:)

translation/japanese/patterns/trusted-committer.md
```diff
@@ -1,10 +1,6 @@
-{% hint style="info" %}
-このパターンで紹介されている "**Trusted Committer**" という語は、"**信頼されたコミッター**" や "**トラステッドコミッター**" と訳されます。本パターンブックでの用語の使い分けに関しては「[トラステッドコミッターというロールと呼称に関して](#trusted_committer)」をご覧ください。
-{% endhint %}
-
 ## Title
 
-信頼されたコミッター (トラステッドコミッター)
+トラステッドコミッター
```
spier, line 3: @yuhattor This looks like you had multiple translations for TC here?

The main one, and then a second one in brackets?

Note that you can use the '## Alias' section for such alternative titles.

spier (approved):
I can not review the Japanese itself but when you are happy with it, so am I ;)

Maybe for future PR could we get a second person from the JP patterns community to review these PRs from a pure language perspective?

Lastly, before merging this PR, please confirm if the failing link check iss as in one of your files.

spier:
@yuhattor please change the PR target to 'main'. All changes need to be made there first and then pushed to 'book-jp' later. 

I am not at my laptop so cannot makes these changes right now. Also excuse my brevity. 

I can help tomorrow to get these changes published in the book.

yuhattor:
@spier 
> I can not review the Japanese itself but when you are happy with it, so am I ;)
> Maybe for future PR could we get a second person from the JP patterns community to review these PRs from a pure language perspective?

Thank you:) Yea.. Our community in Japan is still small, but I hope we will have more contributers soon!

> Lastly, before merging this PR, please confirm if the failing link check iss as in one of your files.
Sure, it looks the error is caused by GitHub Actions sending too many requests, resulting in 429.

I wonder why it failed only this time, looking at the lycheeverse/lychee settings, it looks like the default values regarding retries are applied

```
  ERROR	https://github.com/tapjdey/InnerSource_Project_Fitness
    Too Many Requests (HTTP error 429)
```

Even if I re-run it a few times, it gets rejected due to strict throttling on the GitHub side...

image

yuhattor:
>  please change the PR target to 'main'. All changes need to be made there first and then pushed to 'book-jp' later.
I am not at my laptop so cannot makes these changes right now. Also excuse my brevity.
I can help tomorrow to get these changes published in the book.

Oh, that's right. Thanks for pointing that out!

translation/japanese/patterns/trusted-committer.md
yuhattor: TC is a very important word, and we had a deep discussion in the community on how to express it.
Long story short, we decided not to translate "Trusted Committer" as suggested by ICS. Japanese has katakana, a character that represents a sound.

So we are okay with using one word here, but if we need to alias it in the future for other terms, we will do so

yuhattor:
Regarding the rate limit, it looks like I found a way to avoid it.

> GitHub Token
To avoid getting rate-limited while checking GitHub links, you can optionally set an environment variable with your Github token like so GITHUB_TOKEN=xxxx, or use the --github-token CLI option. It can also be set in the config file. [Here is an example config file](https://github.com/lycheeverse/lychee/blob/master/lychee.example.toml).

https://github.com/lycheeverse/lychee

spier:
I believe we are already using they. 
See `.github/workflows/link-checker.yml`

After adding that the GitHub rate limiting issues were resolved. However apparently not for good. 

That token is generated by GHA on every CI run. 
I suspect it has the same rate limits as a personal access token would have. Not sure though. 

I say let's merge this PR into main, and watch how the link checker behaves in the future. 



> Regarding the rate limit, it looks like I found a way to avoid it.
> 
> 
> 
> > GitHub Token
> 
> To avoid getting rate-limited while checking GitHub links, you can optionally set an environment variable with your Github token like so GITHUB_TOKEN=xxxx, or use the --github-token CLI option. It can also be set in the config file. [Here is an example config file](https://github.com/lycheeverse/lychee/blob/master/lychee.example.toml).
> 
> 
> 
> https://github.com/lycheeverse/lychee
> 
>

translation/japanese/patterns/trusted-committer.md
spier: One day you need to tell me more about what "katakana" means. Sounds great for sure :)

But for now, let's resolve this thread.

spier (approved):

spier:
@yuhattor I got access to a proper laptop again.

Changed the base branch of this PR to `main`, and approved the PR.

If this all looks good to you, please merge.

I can then help tomorrow to get this released (i.e. pushed from `main` to `book-jp`.

yuhattor:
Thank you for coordinating pull requests!!
Also, thanks for the comment about rate limits as well.
Yes, let's merge them once:)

PR #446, opened 2022-09-01
Add Airbus as known instance of the InnerSource Portal pattern

sicot-f:
Airbus deployed the SAP InnerSource portal on their GitHub Enterprise platform.


spier:
Nice, thank for contributing another instance of this pattern in the wild @sicot-f!

You will see Airbus showing up shortly in the published pattern at:
https://patterns.innersourcecommons.org/p/innersource-portal#known-instances

Issue #430, opened 2022-07-25
Japanese Patterns Book - Ideas for further improvement


spier:
As part fo #422 we identified some things that are not essential for the first release of the Japanese Patterns Book.
Still they could be considered to improve the book further.

These ideas are listed below, and after the first book release we can decide how to prioritize these.

## Future Work for the Japanese Patterns Book

- maintenance - The `lint-patterns.yml` GHA checks that all patterns contain the mandatory headlines as defined in `.meta/pattern-template.md`.  This isn't essential for JP patterns, as those are translations of EN patterns anyways, for which we have confirmed that they have these mandatory headlines. We can try to adapt this GHA to also check the required headlines in JP patterns but it isn't essential
- ✅ structure/maintenance - I am suggesting to rename the folder `translation/japanese` to `translation/ja` to make it consistent with the book URL as well as the `book/ja` folder - see #452 
- structure/maintenance - inside of `translations/jp` keep the same folder structure as in `patterns` (that way all links to other patterns and assets will stay the same). => As this would require to check all links in the JP patterns again I am pushing this to Future Work
- gitbook publishing - update `book-jp` branch automatically when changes to `main` are made. => I am opting to do this manually for now, as it does not require more than merging `main` into `book-jp`. In the future we can automate this, maybe with this [action](https://github.com/repo-sync/pull-request) to create a PR and [this one](https://github.com/marketplace/actions/enable-pull-request-automerge) to auto-merge. or better to merge without a PR?
- content - what to do with the mindmap? `book/explore-patterns.md` - We decided to keep it in English for the 1st release
- content - the captions of most images are still in English e.g. see [this](https://innersourcecommons.gitbook.io/staging/v/jp/p/innersource-portal)
- ✅ content - translate the book cover itself to Japanese too? (I do have the source files, so if somebody helps me with the translations then we can do this)

PR #447, opened 2022-09-03
Moving Japanese content to `translation/jp`

spier:
This PR contributes to a part of #430.

I moved all files from `translation/japanese` to `translation/jp` and adapted scripts and docs that contained the old folder name.

By changing this folder name things are more consistent with branch names (`book-jp`) the other book files (`/book/jp`) etc.

Deployed a staging version of this to:
https://innersourcecommons.gitbook.io/innersource-patterns-staging/v/jp-new-folder/

As part of this I also improved the GHA workflow.
Now any folder starting with `book-jp` will generate the Japanese book, which should be helpful for testing purposes.


spier:
@yuhattor could you review this?

In theory it is a relatively small change but as I am moving/touching all files, I want to make sure that I did not break anything.

yuhattor (commented):
@spier 
Very nice!
Just a little input from the community, but jp represents the country and ja represents the locale. If we ever get a Chinese translation in the future, it might be a good idea to adapt the i18n rules sometime in the future. For now, I think jp is fine.
But I don't want to change it too often from SEO point of view. I thought maybe this could be fixed now. What do you think?

spier:
Oh, good one.

So do you want to replace `jp` with `ja` everywhere?

- branch name => `book-ja`
- book URL => `https://patterns.innersourcecommons.org/v/ja/`
- folder name => `translation/ja`
- ... (maybe we got some wherewhere else too?)

If we want to change it, then we should do it very soon though (SEO / not breaking links).
I would suggest to do this in a new PR though (I can take care of that, if you like).

yuhattor:
I think so. I'll ask the community for some input. We are generally more familiar with jp than ja.
As you said, I think it would be good to do it somewhere different from this PR for once:)

If you think it's good as it is now, please merge it:)

spier:
> [...] We are generally more familiar with jp than ja. [...]

@yuhattor Did you get this the wrong way around possibly?

Earlier I understood that `ja` is the correct language short code to use, right?

> If you think it's good as it is now, please merge it:)

I actually changed my mind on this.

Let's wait until you got input from the Japanese community. Then I can adapt this PR as desired, merge it, and change the gitbook config to build from the correct branch and use the desired URL prefix.

That way we only do this work once, and not twice :)

yuhattor:
Sorry, that was confusing.

Now let's change patterns to ja!
ja is the official i18n term.
This is also how it looks on innersourcecommons.org
https://innersourcecommons.org/ja/learn/learning-path/

The Japanese community is basically trying to comply with Upstream. And ja is the "correct" code.
So we decided to go with ja. Thank you always, I'd appreciate it if you could fix it!

spier:
I renamed the branch from `book-jp-folder-move` to `book-ja-folder-move` in order to get the modified GHA to run.
That automatically closed this PR as well. Did not know that that would happen but makes sense I guess :)

spier:
Continued this work in https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/452

PR #452, opened 2022-09-14
Moving Japanese content to `translation/ja`

spier:
This PR contributes to a part of #430.

I moved all files from `translation/japanese` to `translation/ja` and adapted scripts and docs that contained the old folder name.

By changing this folder name things are more consistent with branch names (`book-ja`) the other book files (`/book/ja`) etc.

Deployed a staging version of this to:
https://innersourcecommons.gitbook.io/innersource-patterns-staging/v/ja-test/

As part of this I also improved the GHA workflow.
Now any folder starting with `book-ja` will generate the Japanese book, which should be helpful for testing purposes.


spier:
@yuhattor could you review this again? Now using `ja` in all the places.

When reviewing this, could you also double check the documentation, to make sure that we are explaining correctly how this all works? Thanks :)

Btw I would already go ahead and change the URL on Production to `/v/ja`, as that change does not require this branch here to be merged. I will confirm this with you in Slack to be sure.

yuhattor:
Thank you! It's amazing:)
I think it is now clearer that it is in Japanese.

yuhattor (approved):
lgtm:)

PR #460, opened 2022-09-27
Correcting typo in document-your-guiding-principles.md

claredillon:
Changing advise to advice


spier:
Oh, it is "to advise so." but "the advice"? Yikes, English is hard :)

Thanks for the fix @claredillon!

spier (approved):

spier:
The fix is live now 🚀 
https://patterns.innersourcecommons.org/p/document-your-guiding-principles

Issue #462, opened 2022-09-27
Improve visuals for Standard Base Documentation


spier:
We want to promote the [Standard Base Documentation](https://patterns.innersourcecommons.org/p/base-documentation) pattern in our upcoming [ISC October newsletter](https://github.com/InnerSourceCommons/InnerSourceMarketing/issues/181).

In preparation for that I am trying to improve the visuals used in the pattern.

For now I have just created a remake of the existing visuals, using standard ISC colors and ready made visuals for characters. We could also try completely new visuals. However that might take more time than we have, as the newsletter is meant to go out early next week.

### Old visuals:

![Screenshot 2022-09-27 at 21 47 13 ](https://user-images.githubusercontent.com/163029/192621559-0b674d55-dc40-4c98-8ec3-066755cb98f8.png)

### Proposal:

![ISC _ Exploring visuals patterns](https://user-images.githubusercontent.com/163029/192621179-2f7d09e1-cd82-4bfb-8b63-d47839633510.jpeg)

spier:
@MaineC anything you wish you could improve about the existing visual for this pattern?

Also, what do you think about the proposal above?
I was thinking that I might split this up into two separate images i.e. one "for users" and one "for contributors. That way we can include the corresponding image next to the explanation of `README.md` and `CONTRIBUTING.md` in the pattern.

MaineC:
The proposal looks great. And about splitting: makes a lot of sense.

spier:
@MaineC the new visuals are live now. Hope you like it :)
https://patterns.innersourcecommons.org/p/base-documentation

PR #463, opened 2022-09-30
Add new images for Standard Base Documentation

spier:
Implements #462


spier:
I take this [comment](https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/462#issuecomment-1260392045) by @MaineC as confirmation that these images are good to be merged.

spier:
I have to keep the old image around, as the Japanese book is still using it.

Also when creating these new images I created a folder based on the filename of the pattern below `assets`, so that all visuals for a pattern are grouped together in that folder. Also allowed me to add a README with license info etc.

Not sure if this approach scales but trying it out.

PR #471, opened 2022-10-14
Link Check just for files in a PR

spier:
Our current link checker will check all links in the main content files (mostly markdown files).
It will do that whenever a PR runs.

With this approach the check reports broken links in files that were not even touched by the given PR.
This can be confusing for the authors of those PRs.

Therefore I am trying something different:
- add a new link check that will only check the links in the files modified in a given PR
- modify our existing link check, to only run on a schedule once per week


spier:
@yuhattor what do you think about this approach to checking links using GitHub Actions?

yuhattor:
It's good to reduce the number of requests. Also, it saves waiting time to get the results:)

PR #474, opened 2022-10-21
Add image visualizing the four governance levels

MaineC:
I'm sure we can find a better visualization from someone who can create images in the InnerSource Commons look and feel. What I want the image to communicate:

* Each level adds more karma to the contributing team.
* The last level though needs a switch in thinking - for the last level what is created in a way is a project on neutral ground where both teams can collaborate each in enlightened self interest. Typically though not all (contributing/host-)team members will be equally active in that project.

What I think would make sense to also capture in the visual but failed to capture myself: Each step requires more transparency and shared communication resources between both teams.

For full disclosure: Feel free to dump this pull request in favor of a different picture with the correct look and feel.

### DIFF
--- a/patterns/1-initial/governance-levels.md
+++ b/patterns/1-initial/governance-levels.md
@@ -58,6 +58,9 @@ Examples of building blocks:
   on the project direction. Everyone has an equal say in who else to add to this
   group.
 
+![image](https://user-images.githubusercontent.com/70953/197144534-42dafb93-9d2d-46d1-a1f2-e7a70e593f6c.png)
+
+
 ## Resulting Context
 
 Teams can adopt InnerSource best practices in a step-by-step way. By documenting
--- a/patterns/1-initial/governance-levels.md
+++ b/patterns/1-initial/governance-levels.md
@@ -60,7 +60,6 @@ Examples of building blocks:
 
 ![image](https://user-images.githubusercontent.com/70953/197144534-42dafb93-9d2d-46d1-a1f2-e7a70e593f6c.png)
 
-
 ## Resulting Context
 
 Teams can adopt InnerSource best practices in a step-by-step way. By documenting


spier:
Embedding the image here as well for faster review.

patterns/1-initial/governance-levels.md
```diff
@@ -58,6 +58,9 @@ Examples of building blocks:
   on the project direction. Everyone has an equal say in who else to add to this
   group.
 
+![image](https://user-images.githubusercontent.com/70953/197144534-42dafb93-9d2d-46d1-a1f2-e7a70e593f6c.png)
+
+
```
spier:
```suggestion
```

spier:
@MaineC just confirming that I understand the image to governance level association correctly:

* top left - Bug reports and issues welcome
* top right - Contributions welcome
* bottom left - Shared write access
* bottom right - Shared ownership

Correct?

MaineC:
Correct

spier:
@MaineC I read the pattern again. 

To help with the visual I need to understand the pattern itself better.

I noticed, that while the title is "Transparent Governance Levels", we don't talk about "levels" in the Solution but rather about "building blocks".

However from your description in the pattern and also from the draft of the visual I get the impression that these levels are really building on top of each other i.e. a project cannot state that they are willing to give the influence level "Contributions welcome", without also giving "Bug reports and issues welcome". Correct?

If that is indeed the case, I would suggest to:

- give them a level number : K1, K2, K3, K4 (K for Karma; or maybe L for Level?)
- state somewhere that every level always gets all permissions of the lower levels (much like you see it in pricing models for online service "this subscription contains all features of lower tiers plus ABC")

I will play around with the visual a bit, to see what I can come up with.
If it is too hard to express everything in one visual, we could consider splitting up the different concerns into multiple visuals.

spier:
@MaineC here an idea that played around with. There are some things that I don't like about it but I figured sharing early is typically the best. Let's see how GitHub renders this one :)

[Update] Found some bugs in my draft already:
- boxes in K1 should have that purple background
- dotted box in K4 should also be purple (as that always indicates the thing added in a new level)
- roadmap image is missing from K4

combined

MaineC:
I like that picture - in particular the hint to the contributor potentially becoming a team member at the bottom.

spier:
Closing this in favor of the other work happening in PR #765 

I believe I have ported all relevant info form here over there.

PR #480, opened 2022-11-03
[docs] Add a chart to explain the levels of the mindmap

spier:
I was reviewing our docs for the mindmap generation in the context of #479.
As part of that I extended the docs a bit, to explain the levels better.


spier:
FYI @fwan2000

PR #481, opened 2022-11-04
Fix mailto syntax

spier:
The last [link check](https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/3392348120) run reported issues. Now trying to fix those.

### DIFF
--- a/.github/workflows/link-checker-prs.yml
+++ b/.github/workflows/link-checker-prs.yml
@@ -39,7 +39,7 @@ jobs:
         name: Link Checker
         uses: lycheeverse/lychee-action@v1.5.2
         with:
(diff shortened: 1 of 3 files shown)


spier:
Current error
> [me@simaos.net](me@simaos.net): Failed: Unreachable mail address: me@simaos.net: Invalid: Email doesn't exist or is syntactically incorrect (status code: ERR)

I don't understand why this email is reported as wrong.

I don't understand either why lychee still checks it, even though I changed the link checker to stop checking emails.

Further digging is required ...

spier:
Looks like I was looking at an outdated GHA run. Now they are all green.
Merging.

PR #492, opened 2022-11-26
Adding WellSky as a Known Instance to the Core Team pattern

spier:
@rrrutledge your [presentation](https://www.youtube.com/watch?v=kgxexjYdhIc) is from 2021 but I only realize now that through that we can also add WellSky as a Known Instance to the Core Team pattern, can't we?

I wrote a little blurb that I extracted from your video but feel free to replace it entirely, if it doesn't make sense.

Also if you have any online reference for how this pattern was used at Nike, we could add that as well.

### DIFF
--- a/patterns/1-initial/incubator-pipeline.md
+++ b/patterns/1-initial/incubator-pipeline.md
@@ -14,7 +14,7 @@ The team managing the shared code library needs a way to allow InnerSource contr
 
 ## Story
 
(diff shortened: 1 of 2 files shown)


rrrutledge (approved):
Thanks, @spier ❗️I don’t remember if there was a public presentation on Core Team for Nike.

spier:
I figure that googling for `"Core Team" Nike` has very little chances of finding what I am looking for. :)

So I will merge this PR as is.

PR #454, opened 2022-09-15
[TEST] Update JA gitbook with latest from `main`

spier:
Goal: Publish the latest changes to the Japanese translation.

_Note:_ Currently we are doing this manually. Eventually we want to automate it. Expect some bumps in the road :)

PR #493, opened 2022-11-27
[TEST]`book-ja` with latest main merged

robtuley:
To help with #454


spier:
Deleting old branch, as the book release for other languages doesn't work well like this anyways.
Will be using a new approach soon, using gitbooks new "monorepos" configuration (see #578).

PR #497, opened 2022-12-04
Create introducing-metrics-in-innersource.md

Trueblueaddie:


spier:
hi @Trueblueaddie. Thanks for adding this pattern.

There is a pattern with the same name here:
https://github.com/Trueblueaddie/InnerSourcePatterns/blob/patch-2/patterns/1-initial/introducing-metrics-in-innersource.md

If yours is a new pattern, I would suggest to pick a different name, so that we can differentiate both patterns.

I would further suggest to move this pattern to folder `patterns/1-initial`. That way we can give this pattern some time to be vetted by the community before it goes live in our online book.

Please let me know if you need help with any of this, I would be happy to.

Trueblueaddie:
I've reviewed the markdownlint and attempted to correct the spacing concerns. I'm not sure exactly what is necessary to be adjusted here.

Trueblueaddie:
@spier thanks for calling that out. I'll see if mine has anything to add the one you shared.

patterns/introducing-metrics-in-innersource.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Resulting Context
+
+There is a generic metrics strategy in place. Teams understand this is needed and they add their own point of view to this.
+
+## Known Instances  
+
+Santander Bank
+
+## Status
```
spier:
```suggestion
## Status
```

patterns/introducing-metrics-in-innersource.md
```diff
@@ -0,0 +1,46 @@
...
+## Authors
+
+Alberto Pérez García-Plaza
+Daniel Izquierdo Cortázar
+Addie Girouard
+
+## Acknowledgements  
+
+Igor Zubiaurre
+
```
spier:
```suggestion
```

spier:
@fioddor based on #526 I understand that we can close this PR here as duplicate, right?

Just saw that this PR mentions two other authors as well. I know Daniel but Alberto is possibly from Santander Bank itself:
Alberto Pérez García-Plaza
Daniel Izquierdo Cortázar

Should we add those two to the "Acknowledgement" on the pattern?

fioddor:
> @fioddor based on #526 I understand that we can close this PR here as duplicate, right?

I guess so. WDYT, @Trueblueaddie?

> Just saw that this PR mentions two other authors as well. I know Daniel but Alberto is possibly from Santander Bank itself:

Alberto is a former bitergian and currently a university professor.

> Should we add those two to the "Acknowledgement" on the pattern?

Yes.

spier:
Added further acknowledgements via bf274fdee1111486de5e3f793d143c748fe6c3a7

spier:
Closing for now. But if anybody re-discovers this issue, feel free to open it again.

PR #502, opened 2022-12-04
Create initiative-keeps-getting-pushed-back.md

Trueblueaddie:


patterns/2-structured/initiative-keeps-getting-pushed-back.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Resulting Context
+
+Step by step, InnerSource strategies and processes are added to the daily work of the projects.
+
+## Known Instances
+
+Santander Bank
+
+## Status
```
spier:
```suggestion
## Status
```

fioddor:
I believe we can merge this with the [Bad Weather For Lift Off](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/patterns/1-initial/bad-weather-for-liftoff.md) pattern.
WDYT @Trueblueaddie, @ggruetter, Wyane DuPont,  @michaeldorner?

NewMexicoKid:
> I believe we can merge this with the [Bad Weather For Lift Off](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/patterns/1-initial/bad-weather-for-liftoff.md) pattern.

I took a look; the *Bad Weather for Lift Off* donut seems to be more focused on the lack of OSS experience hindering the success of InnerSource where this pattern is more generally focused on business priorities applying pressure to projects. There might be some crossover between the forces involved, but the root problems (and their solutions) appear to be different and perhaps worth addressing in separate patterns.

Just my $0.02.

spier:
Closing for now. But if anybody runs into similar issues and re-discovers this issue, feel free to open it again.

PR #504, opened 2022-12-05
Mercado Libre as known instance of the Standard Base Documentation pattern

privero-meli:
Add Mercado Libre as known instance of the Standard Base Documentation pattern


spier (approved):
LGTM

PR #506, opened 2022-12-05
Mercado Libre as known instance of the Communication Tooling pattern

privero-meli:
Add Mercado Libre as known instance of the Communication Tooling pattern


spier:
Thanks for adding this Known Instance @privero-meli!

Do you use the same type of communication channels listed in the pattern, or did your setup require something else? If you got something interesting to add there, feel free to file a new PR for that.

Further in #420 we are working on further improving this pattern.
We are also discussing a visual there that could support the content of the pattern.
Would be great to get your input there too :)

PR #507, opened 2022-12-05
Mercado Libre as known instance of the InnerSource Portal pattern

privero-meli:
Add Mercado Libre as known instance of the InnerSource Portal pattern


spier (approved):
Thanks for sharing this Known Instance!

PR #509, opened 2022-12-06
Deduplicate entry for 'Time Constraints Hamper Inner Source Progress'

spier:
Noticed that we got this pattern listed twice in our main README.
This PR does a minimal change to fix this.

- Adding alias
- Adding in Patlet that we used in the README. Not a true patlet but rather a modification of the Problem section.
- Remove duplicate entry in README


spier:
Trivial change, so I went ahead without PR review.

PR #512, opened 2022-12-21
Move extensions to Structured pattern

Ssukriti:
This PR moves extensions to structured for publishing and adds to  mind map


Ssukriti:
@spier I have moved pattern to structured here, i also updated mindmap in same PR (which I realize now may not be the way to do it? )
The linkcheck will fail as the pattern is not published yet, I am not sure why the mindmap generation failed .

spier (approved):
@Ssukriti the failure in the GHA that generates the mindmap seems to be related to a permission issue. I will just merge this PR to get the pattern out, and then fix any remaining issues in further PRs.

Thanks again for all your work on this.

spier:
@Ssukriti your pattern is live at https://patterns.innersourcecommons.org/p/extensions-for-sustainable-growth, and the generation also work, once it was merged in to main. So all good 🚀

PR #514, opened 2022-12-29
Create the Chinese translation PR (WIP)

WillemJiang:


spier:
I will pull your branch into a new experimental branch, so that I can test without merging into `main` and also without disturbing your work.

Btw isn't `zh` the language code for Chinese? At least that's what I found [here](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).
So if you don't object, I will rename the folder to that instead of `cn`.

spier:
@yuhattor you can leave this one to me.

WillemJiang:
> I will pull your branch into a new experimental branch, so that I can test without merging into `main` and also without disturbing your work.
> 
> Btw isn't `zh` the language code for Chinese? At least that's what I found [here](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). So if you don't object, I will rename the folder to that instead of `cn`.

Yes，`cn` is a country code, we need to use `zh` as the language code instead.

spier:
@WillemJiang I have merged all your work from here into #515. 

If you could branch of from the new `book-zh` branch for any remaining changes, and also push back to that branch, then we can continue all remaining work on that new PR.

I am closing this PR here, so that we can more clearly see which active branch/PR we are working on.

PR #520, opened 2023-01-16
[Chinese translation] Fixed format typo of README.md

WillemJiang:


spier (approved):

spier:
The process for merging changes to the translations is a bit complicated right now.

In short:
- merge the changes into `main`
- then merge `main` into the respective book branch e.g. `book-zh`

With this approach we make sure that the latest version of the content is visible in `main`, and contributors are acknowledged correctly.
At the same time we allow contributors to hit the "Edit on GitHub" button in the translated gitbooks, in order to make small typo changes. We as maintainers then have to change the target branching to `main` before merging the change but that seems manageable. 

I thought I had written up somewhere why we do it this way but I cannot find that right now.

spier:
The fixed version is deployed now. Looks good to me :)
https://patterns.innersourcecommons.org/v/zh/fu-lu/e-wai/readme-template

WillemJiang:
@spier I just submitted the PR through the "Edit on GitHub" on the Gitbook page. 
Can we set up a GitHub Action to cherry-pick the patch into gitbook-zh branch?

spier:
@WillemJiang maybe. 

One other issue that we ran into with our current approach:
`book-zh` is a protected branch, so some automations with GitHub action were failing, as the bot users that are executing the GitHub action did not have the permission to commit to the branch or something similar.

Just so you know:
I am not super happy with how maintaining the translations works in gitbook. That's why I am looking at alternatives to gitbook right now. However this will take a bit longer, as I haven't found a simple solution yet, even when using other tools like mkdocs etc.

PR #527, opened 2023-02-27
Create new pattern: Group Support

rrrutledge:
View [the document](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/rrrutledge-patch-1/patterns/2-structured/support-as-a-group.md).


rrrutledge:
@jeffabailey interested to review.

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,84 @@
+## Title
+
+Support as a Group
+
+## Patlet
+
+What happens when the team or individual that was supporting an InnerSource project is no longer able to do so?
```
jeffabailey:
```suggestion
What happens if a team or individual no longer supports an InnerSource project?
```

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,84 @@
...
+Form a group of interested individuals from anywhere in the company to keep the project going.
```
jeffabailey:
```suggestion
Keep the project alive by forming a group of interested individuals.
```

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Problem
+
+* A popular InnerSource project is orphaned with no one to support it.
```
jeffabailey: Orphaned implies no support.

Alternatively: A popular InnerSource project is unsupported.

```suggestion
* A popular InnerSource project is orphaned.
```

jeffabailey:
> @jeffabailey interested to review.

Added some. I will add more this evening.

rrrutledge:
> Added some

Like you have a review in progress with comments queued up?  Nice!

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,84 @@
...
+When forming, this group should identify or create:
+
+* A chat channel where questions and discussion about the project should happen.
+* Clear contributing instructions.
+* Link to both of the above in the project README.
+* A chat channel just for the [Trusted Committer]s.
+
+The group should do its best to manage these aspects of the project:
+
+* **Maintenance**.  If the project is flat-out broken for the standard use case, then fix it.
```
jeffabailey: It seems like we're defining how to maintain an InnerSource project in this Patlet.

Should we create one specifically for that, and provide guidance on how to convert a production project to InnerSource here?

jeffabailey: I didn't see a Patlet for that in the repo, but I may be missing something.

lenucksi:
Nice document, only addition I would have is that communication and expectation setting should be made absolutely clear that this is "best-effort inner source" and so demands and expectations by others are out of place and "pull request welcome" is the standard answer. Failing to do so and accepting your average corporate "yesterday" deadline will lead to fast burnout of the new maintainer group.

rrrutledge:
Thanks for this feedback, @lenucksi.  Let me look at incorporating that.

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,86 @@
...
+If there is personal outreach for someone to join as a [Trusted Committer], there are generally a number of people that will say "yes".
+Feeling part of a group and being given some structure and responsibility generally motivates people to try their best, which many times ends up being enough.
+
+## Known Instances
+
+* WellSky
+
+## Status
+
+Structured
```
spier: Would you be ok to share this pattern as "Initial" to get started?
That way it won't be published in our online book immediately, and gives us a bit more time to harden the content.

I admit that this is likely a flaw in our maturity process but somehow it makes me nervous to publish a pattern straight from zero into our book.

Would be honestly great to get your thoughts on this topic, as I feel like I am stuck in my own head here.

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,86 @@
...
+## Context
+
+* Popular InnerSource project.
+* No one is actively supporting it.
+
+## Forces
+
+* No one is assigned by their day job to work on it.
+* Everyone is busy.
+* High cost to migrate off the project.
```
spier: This is high costs for the teams that are using this library, right?

rrrutledge:
@jeffabailey let us know if you have some comments/feedback?

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,86 @@
+## Title
+
+Support as a Group
+
+## Patlet
+
+What happens when the team or individual that was supporting an InnerSource project is no longer able to do so?
+Form a group of interested individuals from anywhere in the company to keep the project going.
```
spier: In the Patlet we are looking for problem + solution in two sentences. I would try to focus the 1st sentence more on what the core fo the problem seems to be, which is that the company is not supporting the maintenance of this project.

```suggestion
A heavily used InnerSource project is loosing its maintainers and no new maintainer is assigned by the company.
By forming a group of interested individuals from anywhere in the company you are able to keep the project going.
```

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,86 @@
...
+At first, no one notices, but after a while whenever someone asks "who owns it" there isn't an answer.
+What will happen next?
+Will new teams shy away from using it?
+Will the project stagnate and linger until its users eventually are forced to move on to something else?
+What a shame if that were to happen to a perfectly good and useful project!
+
+## Context
+
+* Popular InnerSource project.
+* No one is actively supporting it.
```
spier:
```suggestion
* No one is actively supporting it.
* The company cannot assign a new maintainer to the project because ...
```

Suggesting to add the reason for not assigning a new maintainer here, or maybe a list of possible reasons.

This bit got me confused, as it didn't seem logical to me.
The library has a lot of users (50 projects).
The migration cost away from this library is high.
50 x high migration cost = super high cost :)

So why wouldn't the company assign a new maintainer?
Wouldn't it be in their best interest?

It is entirely possible that I am misunderstanding the scenario/context here.
And if yes that's great, so that we can clarify this :)

patterns/2-structured/support-as-a-group.md
spier: Adding on to this:
We can decide at the very end whether we want to publish this as Initial or Structured.
So don't worry about this too much for now.

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,86 @@
...
+
+* **Maintenance**.  If the project is flat-out broken for the standard use case, then fix it.
+Keep the project up-to-date as the dependencies and frameworks it uses continue to evolve.
+* **Onboarding**.  If someone has a question about how to use the project, make sure they get a reasonable answer.
+* **Updates**.  If someone wants to add new feature to the project, give them the design and technical support necessary for them to build it so that it both works for them and is a good addition to the project.
+Review incoming pull requests in a timely manner.
+
+Since this group is comprised of volunteers, it is important to communicate that support is "best effort" only.
+Accordingly this model of support is not well-suited for run-time critical, production projects like live APIs.
+It is better suited for projects that are consumed at build-time like libraries/packages/modules.
```
spier: This limitation seems so convincing, that I would even add something about this to the Context.

Maybe:
* the project is a library/package/module that is integrated into other software components at build-time (see Solution for reasoning)

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Resulting Context
+
+* There is some "best-effort" support for the InnerSource project.
```
spier: How long is the project expected to survive in this form of support?

Is the hope that eventually a new maintainer will be identified by the company?
If yes, then this could be added here.

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,86 @@
...
+* Link to both of the above in the project README.
+* A chat channel just for the [Trusted Committer]s.
+
+The group should do its best to manage these aspects of the project:
+
+* **Maintenance**.  If the project is flat-out broken for the standard use case, then fix it.
+Keep the project up-to-date as the dependencies and frameworks it uses continue to evolve.
+* **Onboarding**.  If someone has a question about how to use the project, make sure they get a reasonable answer.
+* **Updates**.  If someone wants to add new feature to the project, give them the design and technical support necessary for them to build it so that it both works for them and is a good addition to the project.
+Review incoming pull requests in a timely manner.
```
spier: I got the sense from this description that the Support Group won't typically implement requested features, will they? Instead they would try the help the requester to implement it themselves?

Maybe add something like this as well. i.e. what the Support Group won't do, or will unlikely be able to do, given their voluntary nature and "best effort" approach.

spier (commented):
Super interesting @rrrutledge!

The pattern write-up looks great already, in terms of following the pattern template and all.

I left specific comment inline in the pattern.

One general thought that kept coming up when reading the pattern:

Can this model of InnerSource support only work as an absolute exception in a company?

As if this were to become the norm, rather than the exception, I would be worried that it would paint the wrong picture of InnerSource. As in some people might reason "we don't need to find/staff a team for this component. Some InnerSource folks will just pick it up because they like the project". So in that sense that might even lead to an anti-pattern of sort?

Looking forward to the conversation, and again thanks a lot for sharing this!

jeffabailey (changes requested):

jeffabailey:
> @jeffabailey let us know if you have some comments/feedback?

Finished submitting a review.

patterns/2-structured/support-as-a-group.md
spier: I think you meant pattern, rather than Patlet above? (I know, its confusing!)

patterns/2-structured/support-as-a-group.md
```diff
@@ -0,0 +1,86 @@
...
+Call for interested volunteers from anywhere in the company to form a group of [Trusted Committer]s to support the project.
+You may need to reach out to specific individuals based on commit or usage history.
+It is important that there are enough so that the burden on each can be reasonably small.
+
+When forming, this group should identify or create:
+
+* A chat channel where questions and discussion about the project should happen.
+* Clear contributing instructions.
+* Link to both of the above in the project README.
+* A chat channel just for the [Trusted Committer]s.
```
spier: We have some other patterns that are cover some of these points already.

It is certainly fine to repeat key elements in this pattern.
By linking to other existing patterns we can let the readers know that they can go into further depth on those topics.

Some relevant patterns that we could link to are:
- https://patterns.innersourcecommons.org/p/base-documentation
- https://patterns.innersourcecommons.org/p/communication-tooling

fioddor:
It might be difficult to find known InnerSource instances. If so we can argue that this pattern is applied in open source, for example [in Debian](https://wiki.debian.org/Teams).

fioddor:
I wonder if we should extend this pattern to cover Special Interest Groups (SIG). Perhaps it's better to just create the SIG pattern and link it. SIGs have a broader scope than just sharing maintenance effort, but the underlying concept is pretty much the same: _"Alone I can't, with friends yes"_.

spier:
> I wonder if we should extend this pattern to cover Special Interest Groups (SIG). Perhaps it's better to just create the SIG pattern and link it. SIGs have a broader scope than just sharing maintenance effort, but the underlying concept is pretty much the same: "Alone I can't, with friends yes".

Have you seen a SIG-like concept in the context of InnerSource? Would be great to hear more about it :)

And yes, I would keep this separate from the pattern in this PR as well.
This PR seems to be about the maintenance of an individual project.

spier:
@rrrutledge wanted to check if you had any questions about the feedback I left in the review, or if there is anything else that you need to push this forward?

Also while reading other patterns, I found that this one here as some overlap with your "Support as a Group":
https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/patterns/1-initial/explicit-shared-ownership.md

rrrutledge:
> if you had any questions about the feedback I left in the review, or if there is anything else that you need to push this forward?

Just need to spend time going through the feedback and updating.  I will take a look.

patterns/2-structured/support-as-a-group.md
rrrutledge: Yes!

patterns/2-structured/support-as-a-group.md
rrrutledge: It would be great!  Like we've been able to link to [Standard Base Documentation](https://patterns.innersourcecommons.org/p/base-documentation) and
[Communication Tooling](https://patterns.innersourcecommons.org/p/communication-tooling) then we could just link to some pattern that describes the minimum viable support needed for InnerSource.

rrrutledge: It would be great if we had it, but I don't think we should gate on it, either.

patterns/2-structured/support-as-a-group.md
rrrutledge: I've having trouble with putting something in the `Context` that makes sense without having to first understand the `Solution`.  Thinking of leaving things as they are?

patterns/2-structured/support-as-a-group.md
rrrutledge: It could go on for an indefinite amount of time like this.
Maybe they find themselves one day in a context that demands another pattern like [Core Team](https://patterns.innersourcecommons.org/p/core-team).
But who's to say?  Should we outline these future possibilities or just stick to a succinct declaration of this situation?

patterns/2-structured/support-as-a-group.md
rrrutledge: OK we'll see at the end.  Reading the definition of [Maturity Levels](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/meta/contributor-handbook.md#maturity-levels) it does say that we need multiple instances before it can be included as `Structured`.

patterns/2-structured/support-as-a-group.md
rrrutledge: Is there a way to link to related patterns?

patterns/2-structured/support-as-a-group.md
rrrutledge: Good point!  Linked instead.

patterns/2-structured/support-as-a-group.md
rrrutledge: `The group is not expected to implement any new functionality for others.`

patterns/2-structured/support-as-a-group.md
rrrutledge: `The company cannot assign a team to support.`

What are the reasons?  Could be varied.  The point is we are not getting a team for it.

rrrutledge (commented):
Thanks for the feedback!  Have a look!

patterns/2-structured/support-as-a-group.md
spier: Hmm. I am reading this for `Structured`:

> Is validated by at least 1 (one) known instance

For the most part this has meant that patterns with one Known Instance will be moved to Structured. But of course we can make our own choice.

patterns/2-structured/support-as-a-group.md
spier: New Patlet is now:
> What happens if a team or individual no longer supports an InnerSource project?
Keep the project alive by forming a group of interested individuals.

patterns/2-structured/group-support.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Author
+
+[Russell R. Rutledge]
+
+## Acknowledgements
+
+[Russell R. Rutledge]: https://github.com/rrrutledge
+[Standard Base Documentation]: https://patterns.innersourcecommons.org/p/base-documentation
+[Communication Tooling]: https://patterns.innersourcecommons.org/p/communication-tooling
```
spier:
```suggestion
[Standard Base Documentation]: ../2-structured/project-setup/base-documentation.md
[Communication Tooling]: ../2-structured/project-setup/communication-tooling.md
[Trusted Committer]: ../2-structured/trusted-committer.md
```

Adding link to Trusted Committer pattern.

Turning all links into relative links to the markdown files, rather than linking to the deployed online book.
These link automatically get set correctly once the book is deployed to gitbook.

patterns/2-structured/support-as-a-group.md
spier: > Is there a way to link to related patterns?

You can link to patterns as part of the regular text in any section, just like you already did.

Often times this happens in the sections
a) "Solution" - think "pattern X is already good, but we needed something else for reasons A, B, C"
b) "Resulting Context" - think "the situation has been improved by implementing this pattern. however as a result another issue may arise, which can be addressed by pattern XYZ"

You could also add a new section `## Related Patterns`.

I favor the former approach. You would have to explain anyways how you see the patterns related, and doing so as part of the common sections of the pattern seems most natural.

pattern-categorization/innersource-program-mind-map.md
```diff
@@ -90,6 +90,8 @@
 
 ##### [Core Team](https://patterns.innersourcecommons.org/p/core-team)
 
+##### [Support as a Group](https://patterns.innersourcecommons.org/p/support-as-a-group)
```
spier:
```suggestion
##### [Group Support](https://patterns.innersourcecommons.org/p/group-support)
```

spier:
@rrrutledge I am curious to hear your thoughts and experience on this angle:

> Can this model of InnerSource support only work as an absolute exception in a company?
> 
> As if this were to become the norm, rather than the exception, I would be worried that it would paint the wrong picture of InnerSource. As in some people might reason "we don't need to find/staff a team for this component. Some InnerSource folks will just pick it up because they like the project". So in that sense that might even lead to an anti-pattern of sort?
> 
> Looking forward to the conversation, and again thanks a lot for sharing this!

Also want to second the possible risk of maintainer burnout, that was raised [here](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/527#issuecomment-1452317673).

If you haven't seen any these risk materialize, or don't share these concerns, then of course we should not add it to the pattern though. This pattern should reflect the reality of how you have experienced it "in action".

patterns/2-structured/support-as-a-group.md
spier: I see. Still confused by this, I have to admit :)

patterns/2-structured/support-as-a-group.md
spier: I would find it relevant to put something into the Context, as it is a key limitation, that readers should compare to their own Context.

i.e. knowing that the pattern is meant to be applied to build-time dependencies should be highlighted early.

What were your concern with the proposal above?

patterns/2-structured/support-as-a-group.md
rrrutledge: Oh I see.  Only one required for structured.  I think that this is why I submitted this as `Structured` initially - it's something that I am using at WellSky.

patterns/2-structured/support-as-a-group.md
rrrutledge: The company is probably short-staffed in general.

rrrutledge:
> curious to hear your thoughts

I think it's good to note that actually staffing the project is a preferable avenue (e.g. ***Core Team***) for the reasons that folks have brought up.  Is there a **Caveats** or **Disclaimers** section where we can put things like that?

patterns/2-structured/support-as-a-group.md
rrrutledge: I think it's difficult to understand why a build-time dependency is a part of the context without first reading the solution.

rrrutledge: I added a line about it in the **Context**.

spier:
> I think it's good to note that actually staffing the project is a preferable avenue (e.g. _**Core Team**_) for the reasons that folks have brought up. Is there a **Caveats** or **Disclaimers** section where we can put things like that?

@rrrutledge you can create a **Caveat** or **Disclaimers** section in this pattern, if you like. It isn't part of our standard pattern template though.

Personally I would integrate this extra info into the **Resulting Context**, as using this pattern creates new risks in your org. i.e. those risks are a result of using this pattern. One could say sth along the lines of:
> While this pattern can provide a solution for the short to medium term, in the long term it has the following risks <A, B, ... D>. Therefore it should always be the preferable avenue to actually staff the project e.g. by means of a Core Team, or a regular host team that maintains the project.

rrrutledge:
> integrate this extra info into the Resulting Context

Done.

patterns/2-structured/group-support.md
```diff
@@ -0,0 +1,89 @@
...
+Since this group is comprised of volunteers, it is important to communicate that support is "best effort" only.
+Accordingly, this model of support is not well-suited for run-time critical, production projects like live APIs.
+It is better suited for projects that are consumed at build-time like libraries/packages/modules.
+The group is not expected to implement any new functionality for others.
+
+## Resulting Context
+
+* There is some fragile support for the InnerSource project.
+* In the long-term the group support is likely to dissolve again at some point.
+If the project continues in the long run, then use this period of stable group support find a long-lived way to support it (e.g. [Core Team]).
```
spier:
```suggestion
If the project continues in the long run, then use this period of stable group support to find a long-lived way to support it (e.g. [Core Team]).
```

patterns/2-structured/group-support.md
```diff
@@ -0,0 +1,89 @@
...
+Since this group is comprised of volunteers, it is important to communicate that support is "best effort" only.
+Accordingly, this model of support is not well-suited for run-time critical, production projects like live APIs.
+It is better suited for projects that are consumed at build-time like libraries/packages/modules.
+The group is not expected to implement any new functionality for others.
+
+## Resulting Context
+
+* There is some fragile support for the InnerSource project.
+* In the long-term the group support is likely to dissolve again at some point.
+If the project continues in the long run, then use this period of stable group support to find a long-lived way to support it (e.g. [Core Team]).
```
spier:
```suggestion
* In the long-term the group support is likely to dissolve again at some point. If the project continues in the long run, then use this period of stable group support to find a long-lived way to support it (e.g. [Core Team]).
```

Combine both lines, as otherwise gitbook will not format the list correctly (unfortunately).

patterns/2-structured/support-as-a-group.md
spier: Yeah. I think we should stick to the definition as outlined in our Maturity Levels.
That means that the pattern is **Structured**, as it has 1 Known Instance, and also meets all other requirements for that level.

I still think that it might be a flaw in our maturity process, as a single org can publish a pattern straight from zero into our book (with themselves as a known instance).

We (as the patterns working group) should possibly review this in the future but for now let's move on and get this PR merged :)

spier: FYI I will also post something similar as my review comment, so that this does not get lost in this thread.

spier (approved):
Thank you for sharing this pattern with us @rrrutledge!
Also a great win for the ISPO Working Group 👍 

The content of this pattern looks great and I am happy to merge this.

The last point we [discussed](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/527#discussion_r1127072087) was whether the pattern should be marked as **Initial** or **Structured**.

The consideration was whether a single org should be able to publish a pattern straight from zero into our book (with themselves as a known instance). Being able to do that may negatively impact the quality of the pattern, as the pattern has little time to get confirmed and improved by the experience of other orgs applying a similar approach.

However as this is how we have currently defined the Maturity Levels for our pattern, we should go ahead and merge this PR.

We (as the patterns working group) should possibly review this in the future but for now let's move on and get this PR merged :)

Please note that this discussion isn't saying anything about this particular pattern, it is mostly a contemplation about the process in which we publish patterns to our online book in general.

spier:
This pattern is live in our book now!
https://patterns.innersourcecommons.org/p/group-support

rrrutledge:
Woah!  Nice!  Thank you!

PR #533, opened 2023-03-14
Trusted Committer: Added Robert Bosch GmbH as known instance

gruetter:


spier:
Forgot to add my review before merging. So please consider this approved :)

spier:
@gruetter thank you for your contribution.

A question about this part:
> Trusted Committers would be funded for 100 % of their time to be able to focus on this role.

The "100%" is confusing me a bit. Were your TCs part of a regular software development team, and then contributing in their TC capacity to other InnerSource projects that their team depended on?

PR #539, opened 2023-05-12
Removing link to innersourceportal.santander.com

spier:
Our automatic link check was [failing](https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/4956839678/jobs/8867775461) due to a broken link to innersourceportal.santander.com.

@yuhattor could you confirm if the formatting of the text in Japanese is still ok, after removing the link?


spier:
@yuhattor I went ahead and merged this PR, so that the GHA that checks the link no longer reports a failure.

But if you could confirm if the formatting of the text in Japanese is still ok, that would be great.

PR #541, opened 2023-05-23
Use glob to select files to be checked by vale

spier:
The output of the vale spell/style checks is not shown in the PR as expected.
Therefore I am running some experiments here to fix this.


spier:
I am marking this PR as "ready for review" as I want to confirm if the lack of GHA annotations is due to that.

spier:
Reminder to self that while the vale GHA currently runs on most PRs, it doesn't output the information as we might expect see [here](https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/5468308314/jobs/9955764282):

Screenshot 2023-07-05 at 21 54 11

I recall that in this PR here I was trying to make things work better but not sure how far I got.

PR #550, opened 2023-06-20
Update Airbus implementation of the innersource portal pattern

sicot-f:
Airbus switched from SAP portal to Backstage/Bazaar as Backstage is becoming the standard internally

### DIFF
--- a/patterns/2-structured/innersource-portal.md
+++ b/patterns/2-structured/innersource-portal.md
@@ -75,7 +75,7 @@ A [reference implementation](https://github.com/SAP/project-portal-for-innersour
 
 ![Santander InnerSource Portal](../../assets/img/santander_portal.png "Banco Santander InnerSource Portal")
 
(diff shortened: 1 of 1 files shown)


spier:
Thank you for sharing this @sicot-f.

Is there any public description of this approach that we can reference?

Also if Airbus would be willing to share more or even demo your use of the Bazaar plugin, I am sure that there would be many people in the InnerSource communication that would be interested in this.

Lastly another aside: If Airbus wants to get listed [here](https://innersourcecommons.org/stories/) as InnerSource adopters, I could help you to make that happen. Just let me know.

sicot-f:
Thanks @spier for accepting this PR :)
We have no public materials available yet. We're just at the beginning of our InnerSource (and open-source) journey. We're not ready to communicate externally as we need more buy-in internally. It's definitely on our roadmap though and we'll let you know ASAP!

spier:
Thank you @sicot-f. 

In that case I am merging this as is.

You can already see the changes live here:
https://patterns.innersourcecommons.org/p/innersource-portal#known-instances

Looking forward to hearing more stories from Airbus in the future :)

Thanks again for your contribution!

PR #551, opened 2023-06-20
Review GHA for i18n consistency check

spier:
This PR is making minor cosmetic modifications to the new GHA that checks the consistency of the translations.
See file: `.github/workflows/i18n-consistency-checker.yaml`

Main purpose of this PR is for me to understand what the GHA does, and to allow me to ask question to @yuhattor (who created this GHA). That way we have some written documentation that others can refer to later.

### DIFF
--- a/.github/workflows/i18n-consistency-checker.yaml
+++ b/.github/workflows/i18n-consistency-checker.yaml
@@ -1,13 +1,8 @@
 name: i18n Consistency Check
 
 on:
-  pull_request:
-    branches:
-      - master
-    paths:
-      - .github/workflows/i18n-consistency-check.yml
   schedule:
-    - cron: '0 0 1 * *' # run at midnight, on day 1 of the month
+    - cron: '0 0 1 * *'  # run at midnight, on day 1 of the month
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -30,7 +30,8 @@ jobs:
           The following files may have consistency issues with the English version. Please check and update the files.\\n\
           \\n\
           This issue is created when there is an update to content/en. It compares the git update history to let you know what updates are overdue. The issue should be closed when the update is complete.\\n"
-          
+
+          # Loop through all the files in the English directory    
           for file in $(find patterns/2-structured -name '*.md'); do
```
spier: Eventually we would have to loop through files in `patterns/3-validated` as well. So far we have no patterns in that maturity folder yet.

Once we do that, we would have to update this line as well:
```
i18n_filename=$(echo "$file" | sed 's/patterns\/2-structured/translation\/${{matrix.language}}\/patterns/g')
```

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -40,7 +41,7 @@ jobs:
             fi
             echo $file
             echo $i18n_filename
-            # Loop through all the files in the English directory
```
spier, line 43: I moved this comment further up, to where this loop starts.

spier: Also putting another comment here, as I cannot comment on those lines directly.

It looks like rows [54-55 ](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/551/files#diff-5fd6ac5b4777f74c7f0d9b3d0bde3b8f441f8c20b518def9833107d1c753e9e4R54-R55)don't work right now? Probably a copy & paste bug from the similar script for innersourcecommons.org?

Extracting the title from the `.md` files is something that we are already doing when generating the gitbook. There we parse the markdown to extract the title though. If we only need the title here, we might be able to find a simpler way to do it. Nothing major though, as I think that showing the filename in the output is enough.

So in theory we could remove these rows entirely.

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -79,6 +80,7 @@ jobs:
             fi
           done
       - name: Create Issue
+        if: ${{ <expression> }}
```
spier: This is not working as is.

However i was wondering if here we could test for the existence of `issue.md` and only run this entire step if that file exists?

spier (commented):
@yuhattor left a couple of questions inline in this PR.

Note that I only read this script once, so I only have a rudimentary idea so far about how it works.

yuhattor:
Sorry for being late to reply. I'm starting to review now! Thank you for your PR!

.github/workflows/i18n-consistency-checker.yaml
yuhattor: I think this title extraction is a one-liner that we are getting to put on the Issue for a list of patterns that need to be edited, so it could not be any shorter. 
But this could be made better.
I'll just fork it over a bit and try to change things in my environment as well.

.github/workflows/i18n-consistency-checker.yaml
yuhattor: That implementation should be in place below, but it doesn't seem to be working.
If there were a similar issue, it would update the new edits as a comment to that issue, so that a new issue would not be created.

Also, since the script is doing the checking for the issue's existence here, there should be an Action in the previous section to make that determination. The process would be like splitting the existing code.

```
            if expr "$existing_issue_id" : '^[0-9]*$' >/dev/null; then
              gh issue comment "$existing_issue_id" -F issue.md -R $GITHUB_REPOSITORY
            else
              gh issue create -t "$issue_title" -F issue.md -R $GITHUB_REPOSITORY -l documentation
            fi
```

yuhattor:
@spier Please check my update on this

So far, it has worked properly.
I also fixed a little script mess.
- Specifically, to describe it in heredoc.
- Now it also skips validated, but I changed it to include it when it loops.
- Fixed Issue creation so that if there is an existing one, it is appended and not a new one is created
  - ↑ use ：jp：expression for emoji instead of Emoji unicode

Here's sample 
https://github.com/yuhattor/InnerSourcePatterns-Japanese/issues/10

yuhattor:
Here's sample 
https://github.com/yuhattor/InnerSourcePatterns-Japanese/issues/10

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -17,88 +12,70 @@ jobs:
...
-              issue+="\\n\`\`\`\\n"
-              issue+="</details>\\n"
-              echo -e "$issue" >> issue.md
-              issue=""
+              result=$(git diff "${parent_hash}" HEAD "$file" | sed '1,4 s/^/# /')
+        
+              # Append to the issue.md file
+              cat <<- EOM >> issue.md
+          <details><summary><b>$content_title</b> ($file)</summary>
+          For more information, please compare [the original file(en)](https://github.com/$GITHUB_REPOSITORY/blob/master/$file) with[the translated file](https://github.com/$GITHUB_REPOSITORY/blob/master/$i18n_filename) . You can view [the differences](https://github.com/yuhattor/innersourcecommons.org/compare/$i18n_last_update_hash...$original_last_update_hash) on GitHub.  The number of days since overdue updates is **$days_since_overdue_updates** days.
```
spier:
```suggestion
          <details><summary><b>$content_title</b> ($file)</summary>
          
          For more information, please compare [the original file(en)](https://github.com/$GITHUB_REPOSITORY/blob/master/$file) with [the translated file](https://github.com/$GITHUB_REPOSITORY/blob/master/$i18n_filename). You can view [the differences](https://github.com/$GITHUB_REPOSITORY/compare/$i18n_last_update_hash...$original_last_update_hash) on GitHub. The number of days since overdue updates is **$days_since_overdue_updates** days.
```

(1)
Looks like we need to add a newline here, so that GitHub can figure out what to interpret as HTML (the `<details>...` line), and what to interpret as markdown (the rest).

(2)
Tried to fix the compare URL. Not sure if I got it right?

(3)
Also fixed some whitespace.

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -17,88 +12,70 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: '0'
-      - name: Check consistency
+      - name: Check consistency and create issue
         id: check-consistency
+        env:
+          GH_TOKEN: ${{ github.token }}
```
spier: Is this token required by the gh cli? `gh issue list`

Was wondering if this is strictly necessary, as I thought that the `GITHUB_TOKEN` that is generated for every workflow run would be available in the env automatically? May be mistaken though but would like to learn how this really works in GHAs :)

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -17,88 +12,70 @@ jobs:
...
-          The following files may have consistency issues with the English version. Please check and update the files.\\n\
-          \\n\
-          This issue is created when there is an update to content/en. It compares the git update history to let you know what updates are overdue. The issue should be closed when the update is complete.\\n"
-          
-          for file in $(find patterns/2-structured -name '*.md'); do
-          
-            # Get the translated file name and check if it exists
-            i18n_filename=$(echo "$file" | sed 's/patterns\/2-structured/translation\/${{matrix.language}}\/patterns/g')
+          # Declare the flags
+          declare -A flags=( ["ja"]=":jp: Japanese" ["zh"]=":cn: Chinese")
```
spier, line 27: Where to lookup the icons to use here?

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -17,88 +12,70 @@ jobs:
...
+        
+          issue_title="${flags['${{matrix.language}}']}: Content Consistency Issue"
+        
+          # Heredoc for issue header
+          cat <<- EOM > issue.md
+          # i18n Contents Consistency Issue
+        
+          The following files may have consistency issues with the English version. Please check and update the files.
+        
+          This issue is created when there is an update to content/en. It compares the git update history to let you know what updates are overdue. The issue should be closed when the update is complete.
```
spier:
```suggestion
          This issue is created when any of the English patterns have changed (in folder `patterns/`). It compares the git update history to let you know what updates are overdue. The issue should be closed when the update is complete.
```

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -1,13 +1,8 @@
-name: i18n check consistency and notify on GitHub Issues
+name: i18n Consistency Check
```
spier, line 6:
```suggestion
# -------------------------
# To add a new language, do the following:
# - add the 2-letter language code in matrix.language
# - add the flag and language name that you want to use at the beginning of the #check-consistency step
# -------------------------
name: i18n Consistency Check
```

Is this description correct?
Anything missing?

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -17,88 +12,70 @@ jobs:
...
-             
-            # print the last updated date of the original file and the translated file
-            
-            # Check if the translated file is updated after the original file
-            if [[ $(($original_updated_at - $i18n_content_updated_at)) -ge 1 ]]; then
-              # Get the title of the content
-              content_header=$(echo "$(cat "$file")" | grep -E '^title+' | sort -r | head -n1)
+        
+            if [[ $((original_updated_at - i18n_content_updated_at)) -ge 1 ]]; then
+              content_header=$(grep -E '^title+' "$file" | sort -r | head -n1)
```
spier, line 53: Doesn't look like the title extract works right now

However let's leave this in here, as we may be adding a frontmatter to all patterns anyways, at which point this should work :)

spier (commented):
Thanks for your work on this @yuhattor.

I think I found (and fixed) some bugs.
Also added some more documentation.

yuhattor:
Thank you for the addition, will comment on it soon...!

.github/workflows/i18n-consistency-checker.yaml
yuhattor: It's correct. Thanks! It's essential part...

.github/workflows/i18n-consistency-checker.yaml
yuhattor: I just read the document too, and yes, I agree. I think you are very right. I'd like to see how it behaved, but I don't think this is necessary.

> Important: An action can access the GITHUB_TOKEN through the github.token context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action. As a good security practice, you should always make sure that actions only have the minimum access they require by limiting the permissions granted to the GITHUB_TOKEN. For more information, see "[Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)."

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -17,88 +17,70 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: '0'
-      - name: Check consistency
+      - name: Check consistency and create issue
         id: check-consistency
+        env:
+          GH_TOKEN: ${{ github.token }}
```
yuhattor:
```suggestion
```

It should be this...

.github/workflows/i18n-consistency-checker.yaml
spier: hehe, well the output of [your test](https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/6009925339/job/16300303176) looks pretty clear :)
> gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -17,88 +17,69 @@ jobs:
...
+          existing_issue_id=$(gh issue list -S "is:issue is:open $issue_title" | cut -f1)
+          echo "existing_issue_id: $existing_issue_id"
+          # Create a new issue or comment on the existing one
           if [ -f issue.md ]; then
             if expr "$existing_issue_id" : '^[0-9]*$' >/dev/null; then
               gh issue comment "$existing_issue_id" -F issue.md -R $GITHUB_REPOSITORY
             else
               gh issue create -t "$issue_title" -F issue.md -R $GITHUB_REPOSITORY -l "Type - Translation"
             fi
           fi
```
yuhattor, line 86:
```suggestion
          fi
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

I guess we still need it...
image

spier:
Thank you for working on this @yuhattor!

Please approve the PR when it looks good to go from your side.
You can also merge it yourself if you like :) (assuming that that is possible)

.github/workflows/i18n-consistency-checker.yaml
yuhattor: No wonder this thing doesn't work.
I was getting the title from the "title: foobar" meta info to match hugo's md style... But this time it's gitbook, so I need to extract from ## Title.
I didn't realize this when I ported...

I've changed it to use awk to extract it, so it should work now!

image

yuhattor:
Thank you! After some aditions... Now I confirmed that it works properly. 
I will merge this now.

Thank you @spier for the contribution!!

.github/workflows/i18n-consistency-checker.yaml
spier: Very pretty!

.github/workflows/i18n-consistency-checker.yaml
spier: @yuhattor this is the only question remaining that I think we should add to the documentation. 

If you could share where to look up the strings like `:jp:`, then I can add that to the documentation.
I might do when working on #582.

PR #559, opened 2023-07-17
Add Known Instance to service-vs-library.md

rrrutledge:


spier (approved):
Thanks you for sharing another evidence of this pattern being implemented in the wild. 

This will be live in the book
shortly after merging!

PR #568, opened 2023-08-09
Add GitHub to release-process as a known instance

zkoppert:
GitHub also uses this release project on several innersource projects. I figure adding to the known instances list builds trust in the pattern so thought I would add to it.


spier:
Thank you for sharing this @zkoppert and for improving the pattern. This will be live shortly at
https://patterns.innersourcecommons.org/p/release-process

Are there any further InnerSource-specific experiences that you at GitHub made with this approach? 
Or anything that you do to motivate your teams to use this approach?

PR #578, opened 2023-08-24
Host translations in gitbook, using monorepo approach

spier:
The current approach for hosting the translations of our patterns is brittle.
As each book needs a custom `.gitbook.yaml` configuration, we have a script that copies the file to the root of the repo, depending on which book we want to render. That in turn also leads to the need for longstanding branches like `book-ja`, `book-zh`, etc.

All of this leads to various issues though, and the deployment of new versions of the book is still largely manual, which effectively means that we rarely do it :( See an example of that here: https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/493

## Requirements

- automatically publish the latest translations, once they are merged into `main`
- don't change any of the URLs in the existing books (don't have to deal with redirects etc) - or a bit softer: change as few URLs as possible

## Solution: monorepo configuration

In the meantime gitbook has created a configuration option to host multiple gitbooks from [monorepos](https://docs.gitbook.com/product-tour/git-sync/monorepos).

In this PR we are trying out how that works, and if it is an approach that would be better suited for hosting our translations.

How it works:
- create one gitbook space for English and each translation (e.g. we would have 3 spaces right now: English, Japanese, Chinese)
- set the "project directory" of that space to `book/jp` (these folders contain the custom content and configuration files for the specific book - in this example for the Japanese book)
- that project directory contains a `.gitbook.yaml`, which points to the content for the given book
- the pattern files can stay where they are (e.g. for en in `patterns/...` and for the translation in `translations/jp/...`

## TODO

- [x] test on a gitbook collection, isolated from Production
- [x] adapt documentation - e.g. `book/en/README.md`, `translation/README.md`
- [x] adapt automation - `.github/workflows/book.yaml` - on any push to `main`, regenerate the TOCs for all books
- [x] adapt configuration (e.g. branch protection rules are no longer needed for `book-jp`)
- [x] merge this PR (so that we can deploy the book and all translations from `main`, using the new monorepos approach)
- [x] (IMMEDIATELY after merge) deploy to Production (requires a change to config of the git-sync in gitbook. see [here](https://docs.gitbook.com/product-tour/git-sync/monorepos#updating)) - we need to do this immediately because gitbook won't build correctly from the new `main` branch any more, using the old configuration
- [x] merge `main` into ##577 to prepare the monorepo configuration for the new Brazilian Portuguese translation
- [x] cleanup: remove branches `book-*` and PRs/issues that are now obsolete


spier:
In early tests of this, I saw something that looked like a bug:
- ✅ The links in the menu work correctly.
- ❌ However when clicking on the links hosted at /toc (toc.md) the links lead to the .md files in our GitHub repo, rather than to the rendered pages in gitbook.

It looks like the issue somehow resolved itself though.

spier:
Notes for how to configure the gitbook space.

* go through all spaces (languages) one by one
  * delete existing git syn
  * reconfigure git sync
    * set branch to `main` (for all spaces)
    * set project directory to `book/en/` for English, `book/ja/` for Japanese etc
  * don't change any other configuration of the space (e.g. Google Analytics or similar). those should be correct as they are.

spier:
Merging to main, so that I can deploy to Production.

PR #585, opened 2023-08-30
Render RFC templates in book

spier:
We already have the README and CONTRIBUTING template embedded directly in our gitbook.
However for some reason we did not do that for the RFC template (referenced in https://patterns.innersourcecommons.org/p/transparent-cross-team-decision-making-using-rfcs).

Changes:
- Adding en RFC template to menu of the book. That way it renders in gitbook, rather than redirecting to github.
- Adding zh translation of RFC template to the book


spier:
This was also a nice quick test if merging this PR will immediately update both the en and zh book.
And it does 🥳 

This is a benefit that we got from using the monorepo approach (see #578).

PR #586, opened 2023-09-01
I18n check: fix empty comment

spier:
I noticed a bug in `.github/workflows/i18n-consistency-checker.yaml`.
When it runs but doesn't find any issues, it will still create a comment to the existing GitHub issue.
Example: https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/583#issuecomment-1701941780

This happened because the script wrote a header to `issue.md`, and later one the existence of the `issue.md` file was used to confirm whether an issue/comment should be created or not. As the header always exists, this was always true.

I fixes this by breaking up the logic into two main blocks:
a) check for inconsistencies between en and translation
b) if inconsistencies were found, create issue/comment

I am sure that my solution could be greatly improved with some more shell scripting knowledge.
Personally I am not super happy with my introduction of yet another file (`issue-full.md`). This could have likely been done differently as well.

Looking forward to any input.


.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -72,18 +61,41 @@ jobs:
...
+      - name: Check issue.md existence
+        id: check_files
+        uses: andstor/file-existence-action@v2
+        with:
+          files: "issue.md"
+      - name: Create issue
+        if: steps.check_files.outputs.files_exists == 'true'
+        run: |
+          # Heredoc for issue header
+          cat <<- EOM > issue-full.md
```
spier, line 96: Rather than introducing a new file, we could also try to prepend the header to the content of issue.md.

Logic in pseudo code:
```
issue.md = <header string> + issue.md (generated in previous steps)
```

Just not sure how to do this in a shell script :)

.github/workflows/i18n-consistency-checker.yaml
```diff
@@ -72,18 +61,41 @@ jobs:
           EOM
             fi
           done
+
+      - name: Check issue.md existence
```
spier, line 84: Not sure if a GHA is needed to check for file existence.

Maybe this can be done directly with a GHA **expression**? Could not find the syntax for that though.
Also assumed that the file-existence-action probably was created for a reason (i.e. that it was not possible differently).

spier:
@yuhattor I left this in draft because I think the code can be improved.

Also I found it a bit hard to test it end-to-end. I did some basic simulations but cannot say that I am 100% confident that this works as expected.

spier:
@yuhattor would be great if you could help me check this PR.

yuhattor:
@spier 
Thank you for the PR and sorry for the delay!!
I haven't looked at this for a long time, and I've forgotten a bit about the implementation of this codebase.
I am forking it and checking it out. Give me a little time.

.github/workflows/i18n-consistency-checker.yaml
yuhattor: This implementation was set up in the past when we needed to do conditional calls on the GitHub Actions side, but as far as the current implementation is concerned, it only does the work when True, so I don't think we need this.

yuhattor: Cross post #636 here
I edited this. Please merge this 🙏 
I think my change is what you meant here. I did not push directly because I am not a bit confident about it, but I used PR!

spier:
I am closing all PRs that are older than 1 year as unmerged.
They are labeled as "stale", and can easily be opened again if the original authors engage again.

To the PR authors:
Please don't take this as disrespect to your original work in any way! We just have tidy up code/issues/PRs every once in a while, to keep the repository and the surrounding community in a healthy and maintainable state. Thank you all!

PR #588, opened 2023-09-12
Adding ADI to innersource-customer-interview-questions.md

kschueths:
Added ADI as known instance


spier (approved):
That's awesome, thanks for sharing this!

Issue #565, opened 2023-08-05
Insights into the growth and maintenance of our Patterns collection


spier:
Our InnerSource Patterns are maintained by a working group within the InnerSource Commons, often called Patterns WG for short. That Patterns WG is creating a report about anything that has happened related to our patterns once a quarter, to give the [Board](https://innersourcecommons.org/about/board/) of the InnerSource Commons visibility into what is going on. These reports have also prove valuable for the Patterns WG itself as it motivates us to pause and look back over the last couple of months every once in a while :)

As an example, see the [2023-05 report](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/meta/boardreports/2023-05.md).

In these reports we also include some measurable things like 
- how many new patterns were proposed (or published in our book)
- how many patterns were improved? (like where did we discover further organizations that are using this pattern)
- etc

However as some of these measures were hard to collect, we did not do this consistently in our reports.

Now I read about the [Issue Metrics GitHub Action](https://github.com/github/issue-metrics), and was wondering if we could use that to generate some of the metrics from our reports automatically? See also this related blog post about [data-driven insights](https://github.blog/2023-07-19-metrics-for-issues-pull-requests-and-discussions/) by GitHub.

spier:
@zkoppert any recommendations about a useful test run that we could do to find out if the [Issue Metrics GitHub Action](https://github.com/github/issue-metrics) would be helpful for the Patterns WG?

As preparation for a test run, **defining the goal** seems like a good activity :)

This could include:

- create a list of metrics that we have tracked in previous reports of the Patterns WG to the board
- (bonus) add some metrics related to the maintenance of our repo that could be helpful to understand how we as maintainers can better support our contributors

Then we could try to generate some of these metrics, using the issue metrics GHA.

Thoughts?

zkoppert:
I totally agree that first finding a goal is great! Listing those out as the top priorities and then finding tools to support that is best. For example, "Goals: Be responsive (respond to pr and issues in a few days or less), 3 pattern improvements every quarter, etc.

Happy to help automate the metrics part once there are clear goals!

spier:
@zkoppert I might work on this as part of `hacktoberfest`. Would you have a bit of time to help me on this?

zkoppert:
Yeah! Definitely!

spier:
Awesome @zkoppert. 

[...]

Issue #581, opened 2023-08-29
Unexpected failure in GitHub Action run


spier:
In a recent pull request (https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/579), one of our GitHub Action (GHA) workflow runs failed.

Overview of failed GHAs:
![Screenshot 2023-08-29 at 09 49 14](https://github.com/InnerSourceCommons/InnerSourcePatterns/assets/163029/c7a9cd2f-7fe2-4b3d-9ced-f454527a0272)

Now when looking at the [details of that workflow run](https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/6003953262), I noticed something interesting:
* when you click on "workflow file", it takes you to the `book.yml` file from this branch/PR (which comes from a fork)
* however the file that really got executed is actually [the one](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/.github/workflows/book.yml) on the `main` branch of the upstream. I am so sure about this as the main branch contains the config for a matrix build for this GHA, while the branch from the PR does not have that matrix build yet)

So at the very least I am starting to wonder how GHAs really work. Seems like I might have some misunderstandings here.

My questions are:
- which GHAs are executed? the ones on the `main` branch, or the ones on the PR branch, or both?
- `main` and PR contain different versions of the same GHA. what happens in that case?

And lastly:
Why does this GHA run fail at all?

It seems like it is looking for a git ref `hive-mind-pattern-1` and cannot find it? Possibly because that ref only exists on the fork but not on the upstream?

Do we have to configure the `actions/checkout` differently, so that it can find that git ref?
Right now we use `ref: ${{ github.head_ref }}`, which is defined as:
> The head_ref or source branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.

I am suspecting that we are checking out the upstream repo `InnerSourceCommons/InnerSourcePatterns` but then trying to point to a ref that does not exist on that upsream.

Just not sure how one would configure this correctly?

spier:
@zkoppert and @yuhattor you are my best resources on GitHub Actions :) Maybe you have a hunch what might be going on here? Would be great to get your help.

zkoppert:
👋🏻 Yeah this is a confusing topic 😓 

For the question, "which GHAs are executed? the ones on the main branch, or the ones on the PR branch, or both?"
[...]

PR #593, opened 2023-10-09
Create pattern-metrics.yaml

zkoppert:
Getting started measuring pattern metrics. Related to #565. 

This pull request implements
- Time to first response metrics for new issues (excluding codeowners created) recurring every month
- Time to first response metrics for new pull requests (excluding codeowners created) recurring every month

cc/ @spier

### DIFF
--- a/.github/workflows/pattern-metrics.yaml
+++ b/.github/workflows/pattern-metrics.yaml
@@ -30,6 +30,10 @@ jobs:
 
     - name: Check out the code so we can get the CODEOWNERS names
       uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            .github/CODEOWNERS
+          sparse-checkout-cone-mode: false
 
     - name: Get usernames from CODEOWNERS file
       shell: bash
--- a/.github/workflows/pattern-metrics.yaml
+++ b/.github/workflows/pattern-metrics.yaml
@@ -1,0 +1,120 @@
+name: Monthly issue metrics
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '3 2 1 * *'
+
+permissions:
+  issues: write
+  pull-requests: read
+
+jobs:
+  issue-metrics:
+    name: issue metrics
+    runs-on: ubuntu-latest
+
+    steps:
+
+    - name: Get dates for last month
+      shell: bash
+      run: |
+        # Calculate the first day of the previous month
+        first_day=$(date -d "last month" +%Y-%m-01)
+
+        # Calculate the last day of the previous month
+        last_day=$(date -d "$first_day +1 month -1 day" +%Y-%m-%d)
+
+        #Set an environment variable with the date range
+        echo "$first_day..$last_day"
+        echo "last_month=$first_day..$last_day" >> "$GITHUB_ENV"
+
+    - name: Check out the code so we can get the CODEOWNERS names
+      uses: actions/checkout@v4
+
+    - name: Get usernames from CODEOWNERS file
+      shell: bash
+      run: |
+        # open a file called CODEOWNERS and load the content into a variable
+        CODEOWNERS_FILE=$(cat .github/CODEOWNERS)
+        
+        # Extract words from CODEOWNERS that start with @ and convert them into a comma-separated string
+        # ie, "-author:spier -author:zkoppert"
+        # This will be used to filter out these codeowners from certain stats in later steps
+        CODEOWNERS_FILTER=$(echo $CODEOWNERS_FILE | grep -o "@[a-zA-Z0-9\-]*" | sed 's/@/-author:/g' | tr '\n' ' ')
+
+        # Print usernames to terminal for easy debugging
+        echo "CODEOWNERS_FILTER: $CODEOWNERS_FILTER"
+
+        # Store CODEOWNERS_FILTERto GitHub Action environment (not permanent)
+        echo "CODEOWNERS_FILTER=$CODEOWNERS_FILTER" >> "$GITHUB_ENV"
+
+    - name: Run issue-metrics tool
+      uses: github/issue-metrics@v2
+      env:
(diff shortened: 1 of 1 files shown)


zkoppert:
I did some local testing of the parts and pieces in these workflows and it does seem to work. Next I'll work on getting metrics on, 
- number of new Structured patterns
- number of new Initial patterns
- number of new Known Instances for an existing pattern

zkoppert:
@spier Any idea what happened on the [failing GitBook Generation](https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/6461754152/job/17542073991?pr=593#step:2:51)? Maybe a permissions issue?

spier:
@zkoppert the issue is likely still related to issue #581. 
For the time being we can just ignore those failing Gitbook generations.

spier:
@zkoppert thank you for this PR!

I was going through your code, trying to understand how it works, and what it does.

Where I got stuck was that I don't see the `issue_metrics.md` file mentioned in the GHA in this PR.
Is it missing? Or does it get created by something else that I don't see here?

Thanks again for helping us collect helpful metrics about the way that contributions work on this repo.

.github/workflows/pattern-metrics.yaml
```diff
@@ -0,0 +1,120 @@
...
+
+        # Calculate the last day of the previous month
+        last_day=$(date -d "$first_day +1 month -1 day" +%Y-%m-%d)
+
+        #Set an environment variable with the date range
+        echo "$first_day..$last_day"
+        echo "last_month=$first_day..$last_day" >> "$GITHUB_ENV"
+
+    - name: Check out the code so we can get the CODEOWNERS names
+      uses: actions/checkout@v4
```
spier, line 32:
```suggestion
      uses: actions/checkout@v4
        with:
          sparse-checkout: |
            .github/CODEOWNERS
          sparse-checkout-cone-mode: false
```

I guess it is fine, in terms of performance, to checkout the whole repo?

We could only checkout the `.github/CODEOWNERS` file, if that is the only think we need? Of course then we would have to change that again, if we ever want to use any other files from the repo.

Code above based on [this](https://github.com/actions/checkout#fetch-only-a-single-file).

spier:
Ah, I think I see it now in the docs:
https://github.com/marketplace/actions/issue-metrics

The `github/issue-metrics@v2` step is creating the `issue_metrics.md` file, right?

.github/workflows/pattern-metrics.yaml
```diff
@@ -0,0 +1,120 @@
...
+        echo "CODEOWNERS_FILTER: $CODEOWNERS_FILTER"
+
+        # Store CODEOWNERS_FILTERto GitHub Action environment (not permanent)
+        echo "CODEOWNERS_FILTER=$CODEOWNERS_FILTER" >> "$GITHUB_ENV"
+
+    - name: Run issue-metrics tool
+      uses: github/issue-metrics@v2
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        SEARCH_QUERY: 'repo:InnerSourceCommons/InnerSourcePatterns is:issue created:${{ env.last_month }} -reason:"not planned" ${{ env.CODEOWNERS_FILTER }}'
```
spier, line 59: Why is `-reason:"not planned"` unsed in here?

zkoppert:
> Ah, I think I see it now in the docs: https://github.com/marketplace/actions/issue-metrics
> 
> The `github/issue-metrics@v2` step is creating the `issue_metrics.md` file, right?

Yup, that is exactly it. The `issue_metrics.md` file gets created by the `github/issue-metrics` action.

spier:
@zkoppert if you like, you can pull from main again. I made the CODEOWNERS file a bit simpler, so you will have less clutter in there when extracting the info from there via your script.

.github/workflows/pattern-metrics.yaml
```diff
@@ -0,0 +1,120 @@
+name: Monthly issue metrics
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '3 2 1 * *'
```
spier, line 5: > “At 02:03 on day-of-month 1.”

Alright, why not :)

spier (approved):
From what I understand the GHA creates one issue for each "report type".

e.g.
* issue (1) "Monthly New Issues Data 2023-10-01..2023-10-31"
* issue (2) "Monthly New Pull Request Data 2023-10-01..2023-10-31" 

I left some other comments inline, mostly to understand better how this works.

I am happy to merge this, so that we can try it out.

Would we run in once manually, so that the first reports are generated immediately?

.github/workflows/pattern-metrics.yaml
zkoppert: That is a good idea!

.github/workflows/pattern-metrics.yaml
zkoppert: If someone opens an issue and closes it as "Not Planned" meaning the work hasn't been done and won't be done in the future. I put that in there since most folks don't want that included in their metrics but we can definitely remove it if you prefer to keep those issues in the mix.
Screenshot 2023-10-10 at 10 57 22 AM

.github/workflows/pattern-metrics.yaml
zkoppert: haha. I picked a random time on the first day of the month. People tend to do things on the hour and jobs get backed up so I like to have mine at odd times to avoid queuing. :)

spier:
@zkoppert should I go ahead and merge + run this GHA already as a first test "in Production"? :)

zkoppert:
@spier yes, that sounds good!

.github/workflows/pattern-metrics.yaml
spier: Fun fact 1: This created a 4 second performance improvement :)

Fun fact 2: I had messed up the indentation int he YAML, which made the GHA fail. Was a quick fix though.

PR #598, opened 2023-10-10
[GHA issue metrics] Make the codeowners uniq

spier:
- Make the list of CODEOWNERS uniq. Mostly a cosmetic change to make the SEARCH_QUERY shorter
- Use sparse-checkout for the 2nd job as well


spier:
FYI @zkoppert

PR #599, opened 2023-10-24
pattern-metrics: Merge to one report

zkoppert:
Works on one aspect of #565 by merging the 2 issues generated by this metrics workflow into a single issue.

@spier Let me know what you think of this approach.

The issue generated will look like this:

< Report on issues content >

---

< Report on pull requests content>

### DIFF
--- a/.github/workflows/pattern-metrics.yaml
+++ b/.github/workflows/pattern-metrics.yaml
@@ -82,7 +82,7 @@ jobs:
     - name: Create issue
       uses: peter-evans/create-issue-from-file@v4
       with:
-        title: Monthly New Pull Request Data ${{ env.last_month }}
+        title: Monthly Metrics: New Issues + New Pull Requests ${{ env.last_month }}
         token: ${{ secrets.GITHUB_TOKEN }}
         content-filepath: ./all.md
         assignees: spier
--- a/.github/workflows/pattern-metrics.yaml
+++ b/.github/workflows/pattern-metrics.yaml
@@ -86,3 +86,5 @@ jobs:
         token: ${{ secrets.GITHUB_TOKEN }}
         content-filepath: ./all.md
         assignees: spier
+        labels: |
+            community metrics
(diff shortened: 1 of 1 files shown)


.github/workflows/pattern-metrics.yaml
```diff
@@ -60,69 +60,29 @@ jobs:
...
+    - name: Merge files for report
+      shell: bash
+      run: |
+        echo $'\n\n---\n\n' > hr.md
+        cat ready_to_merge_issues_report.md hr.md issue_metrics.md > all.md
+
     - name: Create issue
       uses: peter-evans/create-issue-from-file@v4
       with:
         title: Monthly New Pull Request Data ${{ env.last_month }}
```
spier:
```suggestion
        title: Monthly Metrics: New Issues + New Pull Requests ${{ env.last_month }}
```

Changing the title of the created issue.

We can really only tell from the title of the issue which reports we are seeing, and in which order.
(well the SEARCH_QUERY printed at the bottom also gives some clues.)

.github/workflows/pattern-metrics.yaml
```diff
@@ -60,69 +60,29 @@ jobs:
...
         token: ${{ secrets.GITHUB_TOKEN }}
-        content-filepath: ./issue_metrics.md
+        content-filepath: ./all.md
         assignees: spier
```
spier, line 88:
```suggestion
        assignees: spier
        labels: |
            community metrics
```

Adding a label to the issue, so that we can filter for them more easily later on.

spier (approved):
Looks great.

Making some inline changes, and will merge and test this shortly.

spier:
Thank you for this contribution @zkoppert !

I made some fixes to the YAML, and now it all works.
Note to self: Don't make un-tested YAML changes directly via the UI. Too easy to mess up indentation :)

Now the created issue with the combined reports looks like this:
https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/601

Based on how this merging of the reports into a single reports works, I had some feature ideas for the issue-metrics GHA itself. Will create a new issue for this.

PR #579, opened 2023-08-24
[Pattern Draft] Hive Mind

mishari:
Initial check in of Hive Mind pattern as discusses in issue #371

cc @spier

PR #608, opened 2023-10-31
Switch to pull_request_target strategy

### LINKED ISSUE #581: Unexpected failure in GitHub Action run

zkoppert:
fixes #581

This changes from running the gitbook workflow on the upstream repo to actually running it on the fork that a contributor has created with a changeset. The hope is that this will address the failing builds described in #581 . In order to test this change I believe we need to merge this through to the main branch and then open a pull request from a fork which would trigger this github action job.

cc/ @spier

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - main
-  pull_request:
+  pull_request_target:
     branches:
       - main
 


spier (approved):

spier:
Thank you @zkoppert. This PR is merged.

Could you also create PR from your fork, to test if this works?
I am worried that if I test this from my own fork, my elevated permissions on the upstream may create misleading results.

spier:
I forgot that I can push changes to a PR submitted from a fork myself.
And through that test the `pull_request_target` from a fork.

Just did that. However it looks like the workflow is still failing. See:
https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/579/checks

PR #617, opened 2023-11-17
Generate mindmap for multiple languages

spier:


spier:
Will try this again on a new brach/PR.

Issue #623, opened 2023-11-24
Showcase the organizations and the patterns they use in our book


spier:
Each pattern already has the "Known Instances" section, where organizations can list themselves when they are using a given pattern.

We could show this information a bit more prominently in [our book](https://patterns.innersourcecommons.org/). 

# Implementation ideas

## Single Page

Maybe a single "Adopters" page at the top level, which would then list organizations, and they patterns they use e.g.

* Airbus
   * InnerSource License
   * Standard Base Documentation
   * Maturity Model
   * ...
* Europace
   * Standard Base Documentation
   * Communication Tooling
   * ...
* Bosch
   * InnerSource License
   * Dedicated Community Leader
   * ...
* Nike
   * Core Team
   * ...

## Dedicated pages for each org

It could even be a dedicated page for each organization. Maybe that would be best?
so e.g. `https://patterns.innersourcecommons.org/adopters/europace`

On that pace we would only list the patterns used by Europace.

That would give the respective orgs a page that they can point to if they want to say "we use these InnerSource patterns".
And if adding themselves to the "Known Instances" of a pattern is the only way to do this, then that might motivate further contributions by other companies.

spier:
I believe we did discuss somewhere if we could link from our [Stories page](https://innersourcecommons.org/stories/) to the patterns used by a specific org.
Implementing this feature here would allow us to do that.

spier:
@sicot-f curious to hear your thoughts on this feature idea.

Currently listing all patterns in use at an organization is only possible like this via a search like this:
https://patterns.innersourcecommons.org/p/innersource-license?q=airbus

Given how the search results are presented in our book (we use gitbook to render the book), it isn't super easy to see all patterns in use at the given org.

Therefore I am wondering if organizations (like "Airbus"), if they could quickly point to all InnerSource patterns that they have adopted, via a dedicated URL similar to `https://patterns.innersourcecommons.org/adopters/airbus`

![Screenshot 2023-11-24 at 17 26 15](https://github.com/InnerSourceCommons/InnerSourcePatterns/assets/163029/a8123b34-fd76-42b7-afe7-d48009821e99)

spier:
I implemented an example of this, to see how it might look & feel like:
https://innersourcecommons.gitbook.io/innersource-patterns-staging/v/adopters-test/adopters/airbus

[...]

PR #624, opened 2023-11-25
Generate landing pages for pattern adopters

spier:
Implements #623

1st implementation approachs:

- take a list of predefinded organization names (e.g. from a file `adopters.txt`)
- loop through all pattern files
- extrac the "Known Instances" section
- check if it contains the name of the organization, and if yes, story it in a list dict[org] => [list of patterns]
- then generate a file `adopters/<org_name>.md`
- write a list of patterns to that file
- commit these changes

- have the `book.yaml` GHA run after this, so that it can pick up these `adopters/*.md` files, and add them to `toc.md`


spier:
Scenario: create a page with subpages in gitbook like this

```
* Parent Page
  * [S1](sub-page1.md)
  * [S2](sub-page1.md)
```

If you don't provide a file for "Parent Page", then gitbook will automatically create a page, where it lists all sub pages. That's kind of nice.

I would like to add some explanation to that parent page though, i.e. to describe what the content on the sub pages is. However I don't know to do that while still keeping the automatic "list all subpages" feature.

spier:
As part of this I am also experimenting with more visual configuration options of gitbook (see `README.md`).
Not sure yet if this will be part of the finale PR.

gruetter:
Would you like updates to the adopters pages for specific companies at this point, @spier ?

spier:
> Would you like updates to the adopters pages for specific companies at this point, @spier ?

hi @gruetter. I am still experimenting with this, so I would say let's finish a basic v1 of this that we can launch, and then add content/features.

This raises an interesting point though:
Do you mean that you want to be able to customise an individual adopter page? e.g. add some custom content to the Bosch page? If we allow that, this would come with some technical challenges (which I am sure we can fix). Also raises questions about how to review such content, not letting it become purely promotional, etc.

Before getting too worried about possible challenges, can you elaborate a bit about the content that you would like to add? Can you share some examples maybe?

PR #633, opened 2023-12-02
Move files from project-setup to standard folders

spier:
Some of our patterns were grouped in a different subfolder `patterns/2-structured/project-setup`.
As we only used this grouping approach once, it created inconsistencies in all sorts of scripts and automations that we use.

Therefore I moved those files from the folder to the `project-setup` folder to the standard folders.


spier:
FYI @MaineC as this change affects patterns that you originally contributed I just wanted to give you an extra ping about this :)

Also left a pointer here, so that people that used direct links to the files in our repo can find the new location of the files:
https://github.com/InnerSourceCommons/InnerSourcePatterns/tree/main/patterns/2-structured/project-setup

PR #635, opened 2023-12-05
Improved Review Committee pattern

### LINKED ISSUE #634: Improvement of Review Committee Pattern
The _Review Committee_ pattern was used at Robert Bosch, but exclusively in the very early stages of InnerSource (2009-2012). After that, we dropped it. I propose to amend the pattern as follows:

- Mention that the pattern was used at Bosch only in the beginning in the Known Instances section
- Add language explaining that the pattern is especially useful when starting your InnerSource journey, and might not be needed later on.

gruetter:
Please see #634 for rationale.

Closes #634


spier (approved):

spier:
@gruetter thank you for the contribution!

spier:
These changes are now live. See:
https://patterns.innersourcecommons.org/p/review-committee#known-instances

PR #637, opened 2023-12-06
Add wiki implementation of InnerSource portal pattern

rrrutledge:


patterns/2-structured/innersource-portal.md
```diff
@@ -78,6 +101,7 @@ A [reference implementation](https://github.com/SAP/project-portal-for-innersour
 * **Airbus** used the [SAP Portal](https://github.com/SAP/project-portal-for-innersource) as a Proof of Concept. It is now using the [Bazaar plugin](https://github.com/backstage/backstage/blob/master/plugins/bazaar/README.md) of [Backstage](https://backstage.io) as the latter became the official developer experience tool internally. It provides a convenient self-registering capability for all the divisions.
 * **Mercado Libre** use an instance of the [SAP portal](https://github.com/SAP/project-portal-for-innersource) to discover existing InnerSource projects within the organization.
 * **Mercedes-Benz** is [using](https://opensource.mercedes-benz.com/news/sponsor_innersource_commonsoss) the SAP reference implementation mentioned above for their InnerSource Portal.
+* **WellSky** has a simple _Confluence_ page were InnerSource and reusable projects are listed.
```
spier:
```suggestion
* **WellSky** has a simple _Confluence Wiki_ page were InnerSource and reusable projects are listed.
```

Suggesting to add this, to make it easier for readers to make the connection between this and the wiki implementation option that you added.

patterns/2-structured/innersource-portal.md
```diff
@@ -52,10 +52,33 @@ Key properties of the portal are:
...
+
+As a simple way to start, you can set aside a page on an internal wiki for listing out available projects.
+An easy way to display this information is in a table with columns giving just a little bit of extra information about the projects.
+Try to have just enough columns so that viewers can determine if they want to learn more about the project, but no more.
+Too much information will make the page overwhelming and difficult to use.
+
+Here is a sample set of columns:
+
+* **Name**.  Name of the project (optionally linked to its homepage).
+* **Brief Description**.
```
spier:
```suggestion
* **Brief Description**.  Explaining the purpose of the project (which problem does it solve?)
```

patterns/2-structured/innersource-portal.md
```diff
@@ -52,10 +52,33 @@ Key properties of the portal are:
...
+* **Technology Pre-requisites**.  You must use these technologies in order to on-board to the project.
+* **Getting Started**.  Link to instructions on how to start using the project.
+* **Chat**.  Link to a chat channel to ask questions about the project.
+* **Sponsoring Team**.  Seeing if a team is behind the project can help others to have the confidence to use it.
```
spier: I have seen "Maintaining Team" and "Host Team" for this in the ISC.

Is there a particular reason why you like "Sponsoring" better here?
e.g. are you worried that the "maintaining" emphasizes a notice of mine vs yours, and with that goes somewhat counter to the core idea of InnerSource?

patterns/2-structured/innersource-portal.md
```diff
@@ -52,10 +52,33 @@ Key properties of the portal are:
 
 When launching the portal, a communications campaign promoting the addition of InnerSource data files or meta-data to code repositories should be considered, to bolster the number of projects displayed within the portal.
 
+### Implementations
+
+
```
spier:
```suggestion
```

spier (changes requested):
Great stuff @rrrutledge!

This makes me want to find orgs that are using Backstage for this purpose, we know they exist, so that we can add this to your new Implementations section :)

patterns/2-structured/innersource-portal.md
```diff
@@ -52,10 +52,32 @@ Key properties of the portal are:
...
 
+### Implementations
+
+#### SAP Project Portal
+
 A [reference implementation](https://github.com/SAP/project-portal-for-innersource) of an InnerSource portal is available on GitHub and open for contributions. It lists all InnerSource projects of an organization in an interactive and easy to use way. Projects can self-register using a dedicated GitHub topic and provide additional metadata.
 
 ![Example of an InnerSource Portal](../../assets/img/portal-overview.png "Example of an InnerSource Portal")
 
+#### Wiki
```
spier, line 63: @rrrutledge I am assuming that this page is manually maintained?
i.e. teams would just come here and add a new row to the table?

Would it be worthwhile to describe somewhere at which point an org might outgrow this solution?
What the benefits of the manual approach are (to get started), and at which point that benefit turns into a problem?

There is room here to elaborate a bit on these points, to help people reason about which implementation to choose for their purposes. Doing this can be really hard, I know, especially for us folks that have looked at this problem space for so long already :)

rrrutledge: Good point - will try to have another “go” to incorporate this feedback.

patterns/2-structured/innersource-portal.md
rrrutledge: Changed to `Host` team.  Yes - maintainer has some connotation that goes against parts of InnerSource.

patterns/2-structured/innersource-portal.md
rrrutledge: OK - added some text.

rrrutledge (commented):
Addressed all feedback.  Thank you!

patterns/2-structured/innersource-portal.md
spier: Looks good!

spier (approved):
LGTM.

Let's deploy this, and see how the readers like it.

Who knows, maybe other patterns could benefit as well from splitting out such "Implementation" information into a separate section in the pattern.

spier:
Note:
This PR also adds a new "Implementation" section to this pattern, which complements the "Solution" and "Known Instances" section.

Will be interesting to see how this benefits our readers. Who knows, maybe other patterns could benefit as well from splitting out such "Implementation" information into a separate section in the pattern.

spier:
@rrrutledge this is live now! https://patterns.innersourcecommons.org/p/innersource-portal#implementations

The formatting of this new content in gitbook looks a bit strange. Apparently gitbook formats the h4 headers the same way as h3. Not a big deal though, let's just leave it as is for now.

PR #644, opened 2024-01-04
Fix Markdown Lint error

rrrutledge:
Add alt text

https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/7386496303/job/20093212394

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 # InnerSource Patterns
 
 <a href="https://patterns.innersourcecommons.org">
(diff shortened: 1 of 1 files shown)


spier:
This is an interesting one. I am not sure where this warning in the screenshot below came from all of a sudden, as I have not seen this before. Anyhow, doesn't hurt to fix it :)

Thanks for doing the fix!

Screenshot 2024-01-04 at 11 37 09

PR #653, opened 2024-02-13
Fix wording in transitioning-contractor-code-to-innersource-model.md

spier:
@zkoppert @claredillon I think there was a double negation introduced here that changed the meaning in the wrong way. At least the description in the "Problem" section that I am changing in this PR did not fit to what the "Patlet" says.

Also while we are talking about this, question to @gyehuda and @zkoppert:
Did you make any new experiences in this space that you can share?

We got another question about this area in Slack, so the topic of "agency code & InnerSource" could be relevant to explore a bit deeper.


zkoppert (approved):
Thanks for fixing! I don't have any new experience in this area to add.

PR #668, opened 2024-03-24
chore: change either to any

jmeridth:
In its current context, there are 3 choices so `any` is more appropriate.  `either` is for binary decisions (between 2 choices).

### DIFF
--- a/patterns/2-structured/common-requirements.md
+++ b/patterns/2-structured/common-requirements.md
@@ -16,7 +16,7 @@ The common code in the shared repository isn't meeting the needs of all the proj
 * Someone (or some project) wrote the code in the first place and contributed it to the repository.
 * The common code is a small percentage of the overall deliverable from any of the projects.
 * Each project has its own delivery schedule, set of deliverables and customers.
-* This pattern applies in either of these situations:
+* This pattern applies in any of these situations:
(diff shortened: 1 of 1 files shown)


spier (approved):
Thank you for the fix. Hope you found it easy enough to figure out how to contribute?

Many of our patterns were written by contributors that speak English as a foreign language. So I am sure there are many more language/grammar improvements like this to be made 😃 

Thanks again!

jmeridth:
@spier Yes.  Contribution process was very easy. Thank you.  I'm re-reading the entire patterns book again and may have other PRs.  This one was the gateway PR and felt very nitpicky but thank you for merging it.  Cheers.

spier:
That's great @jmeridth. Looking forward to any and all contributions that you want to make. 

And no worry about nitpicking, if we really don't want to make a change that you are suggesting, we can just talk about that in any future PRs. No harm done, quite the opposite actually, as it is always great to have somebody with fresh eyes read our patterns again!

spier:
Btw your change is not live in https://patterns.innersourcecommons.org/p/common-requirements yet. Not sure how much delay we should expect there. Will confirm with the gitbook service (which is where we are hosting our book).

PR #680, opened 2024-04-10
Grammatical modifications added to the Galician translation in the texts of the corresponding .md files

raivenra:
Grammatical modifications added to the Galician translation in the texts of the corresponding .md files.


book/gl/toc.md
```diff
@@ -19,29 +19,29 @@ No canto diso edite toc_template.md
...
-* [Trusted commiter](../../translation/gl/patterns/trusted-committer.md) - Moitos proxectos InnerSource recibirán valoracións constantemente ou requirirán a posta en marcha de novas funcionalidades e a corrección de erros por parte dos/as contribuidores/as. Nestas situacións, as persoas encargadas do mantemento do proxecto buscan formas de recoñecer e recompensar o traballo do/a contribuidor/a máis alá das contribucións individuais.
-* [Valoración de proxectos entre equipos](../../translation/gl/patterns/crossteam-project-valuation.md) - É difícil convencer acerca do valor dos proxectos InnerSource desenvolvidos entre equipos que non teñen un impacto directo nos ingresos da empresa. Velaquí unha forma de representar o seu proxecto baseada en datos, que articula o seu valor e amplifícao.
+* [Agradecemento aos/ás participantes](../../translation/gl/patterns/praise-participants.md) - tras unha contribución a un proxecto InnerSource, é importante agradecerlle ao/á colaborador/a polo seu tempo e esforzo. Este modelo brinda unha orientación que non só recoñece dun xeito eficaz a súa contribución, senón que tamén xera un maior compromiso deste/a e doutros/as contribuidores/as.
```
spier: This file cannot be changed directly. The list of patterns here is auto-generated from the patterns files. The summary that you see here is what you see in the "Patlet" section in the respective pattern.

spier:
These changes are now live:
https://patterns.innersourcecommons.org/v/gl

Great job @raivenra and @psanxiao !

spier:
@psanxiao one question:
I noticed that the first letter in the Patlet of the translated patterns was changed to lowercase. I assume that was done to achieve the desired formatting in `toc.md`?

galego-toc

However this also means that when opening the pattern itself, it will look like this. Is this intended and correct in Galician to start a sentence with a lower case letter?

galego-patlet-modifications

psanxiao:
Good point @spier . Changes are made because of TOC, correct. In the TOC, descriptions should be in lowercase in Galician, but as you pointed out, Patlets should start in uppercase since you're right, sentences in Galician start in uppercase.

I guess there is no a good solution for this with the current configuration of the repository. As lesser of two I think it would be better Patlets start in uppercase. Sorry about not realize of that. I have to make some other minor change nexts days so I could take advantage of it and make that change...

Or you think it would be better remove the commit of the PR from main branch and make another one taking this into account?

spier:
I would revert the commit(s) that lowercased the Patlets first, as that will be rather quick. 

And then make the other changes in another PR. 

But up to you.

PR #682, opened 2024-04-12
Upgrade GHAs to node20

spier:
Trying to get rid of these error messages:
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, stefanzweifel/git-auto-commit-action@v4. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.


spier:
As a side effect this will also regenerate the TOC for the Galician translation.

spier:
After merging I pushed one more commit to upgrade further GHAs to node 20:
https://github.com/InnerSourceCommons/InnerSourcePatterns/commit/4d89ae22e027f4fba3cfbff6b32044a39d5736f8

PR #694, opened 2024-06-17
Add GitHub to known Instances

zkoppert:


spier (approved):

spier:
Thank you for contributing to this @zkoppert. 

What would you say is the main difference between a CoC in open source and Inner Source?

zkoppert:
The main difference to me is that employees often have consequences tied to their job performance and compensation for violation where in the open the most you could do is ban a user from contribution/interaction.

spier:
> The main difference to me is that employees often have consequences tied to their job performance and compensation for violation where in the open the most you could do is ban a user from contribution/interaction.

Yeah, that's actually why I was initially thinking that most orgs would not need a CoC in the InnerSource context, as any violations of the norms typically spelled out in a CoC should already be addressed by the org through other means.

My possibly too naive though here was that in a corporate context, people are less likely to misbehave in the ways that we sometimes see in open source, as there is no anonymity, etc.

PR #698, opened 2024-06-24
Fix glossary in book

spier:
The re-generation of the table of contents for the book only works when triggered by a repo Admin.
This is a known issue that we still haven't fixed yet.

Pushing this dummy PR with no changes, so that we can get the glossary published.


spier:
Still had a strange generation error of the GHA that generations the TOC for the book.
I expect it to work on main though.

PR #704, opened 2024-07-11
Fix `calculateScore` function

bn-d:
The bug is accidentally introduced in https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/341 . It incorrectly applied the parenthesis to the function changing `(1 + x / y)` to `((1 + x) / y)`.


spier:
Thank you for this contribution @bn-d!

@Michadelic could you help us with a review on this PR?

spier:
I am just assuming that this is correct, and will merge it ;)

PR #736, opened 2024-12-01
Patterns WG Report / 2024-11

spier:


spier:
I won't have time to make this any better.

So merging as is.

PR #737, opened 2024-12-10
Actions are changed from default branch to version-specified.

taku333:
# Summary
The default branch named `reviewdog` was specified, so it was changed to the version specification.
It is expected to be resistant to errata-ai/vale-action changes.

### DIFF
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -21,7 +21,7 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Vale Linting
-      uses: errata-ai/vale-action@reviewdog
+      uses: errata-ai/vale-action@v2
       with:
         files: '["patterns/2-structured/", "patterns/3-validated/"]'
         vale_flags: "--glob=*.md"


spier:
Thanks for this @taku333!

Back when we introduced this Action it was recommended to use the version tagged `reviewdog`. However that seems to have changed in the meantime, based on the [docs](https://github.com/marketplace/actions/vale-linter). So sounds like a good idea to switch!

Some questions about the implementation:

1. Your PR suggests to just fix the major version, right? i.e. if vale-action upgrades to `2.1.2` or `2.2.x` ... we would still pick these up, as with `errata-ai/vale-action@v2` we would only fix the major version, right?
2. When reading the [docs](https://github.com/marketplace/actions/vale-linter) I noticed that you can also fix the version of vale itself. Currently we are using the default, which means that we will always pull the latest version of vale (currently `3.9.1`). What's your take, should we fix the vale version as well?

And last but not least, just out of curiosity:
How did you come across this repo? Are you part of the InnerSource community in Japan?

In either case, thanks a lot for identifying this issue and for helping to keep our repo stable :)

taku333:
@spier 

Thanks for your reply!

> Your PR suggests to just fix the major version, right? i.e. if vale-action upgrades to 2.1.2 or 2.2.x ... we would still pick these up, as with errata-ai/vale-action@v2 we would only fix the major version, right?

I believe that if the vale-action tag is managed, upgrades below minor will be reflected automatically.

> When reading the docs I noticed that you can also fix the version of vale itself. Currently we are using the default, which means that we will always pull the latest version of vale (currently 3.9.1). What's your take, should we fix the vale version as well?

It may also be necessary to modify the vale-actions.
In that case, I need to modify it again. Shall I prepare the Actions `dependabot.yml` on my end to automate it?
Then, in the future, when a major version is upgraded, a pull request will be sent from dependabot.

> How did you come across this repo? Are you part of the InnerSource community in Japan?

I am not a member of the community.

I found the following book and arrived here.
https://x.com/yuhattor/status/1830489532351107550

I am reading this to see if we can spread the InnerSource culture within the company as much as possible.

spier:
Actually for now let's just go with your current PR.

We have not really had any issues with vale-action or vale in the past, so probably your change in this PR is already making this stable enough.

Thanks again for finding and fixing this!

spier:
> 
> > How did you come across this repo? Are you part of the InnerSource community in Japan?
> 
> I am not a member of the community.
> 
> I found the following book and arrived here. https://x.com/yuhattor/status/1830489532351107550
> 
> I am reading this to see if we can spread the InnerSource culture within the company as much as possible.

@taku333 one more thing: If you like, you can join our [Slack](https://innersourcecommons.org/slack).

There you can find @yuhattor and other folks interested in InnerSource. There is also a channel dedicated to the InnerSource community in Japan, if that should be interesting for you.

Thanks again for the contribution and the nice chat :)

Issue #738, opened 2024-12-17
[Pattern Idea] InnerSource Champion/Ambassadors


rrrutledge:
Have assigned volunteer people drawn from across the organization that are assigned to do a part of supporting InnerSource in their area and across their organization (alongside their regular role in that part of the organization).

dellagustin-sap:
Hello @rrrutledge , there is an entry about this pattern in the discussion, under Pattern Ideas: https://github.com/InnerSourceCommons/InnerSourcePatterns/discussions/729

spier:
In a previous org we had a similar concept but for Hackathons.

In short:
- there was a small central group organizing the global logistics for the hackathon (determining dates. driving global comms into the org. etc)
- then we had volunteer hackathon ambassadors in each offices that were our "eyes and ears" on the ground (reminding people to submit project ideas, organizing local sessions to form teams, organizing food, booking rooms, printing stickers, etc)

This made sense in our org because we had like 8 offices that were spread across the globe. i.e. it was hard for 

Is that idea somewhat similar to what you have in mind here for InnerSource?

Some specific questions to help me think about this:
- What are the problems that an InnerSource ambassador would address that a central group (e.g. ISPO) cannot easily achieve?
- What would be typical tasks of an InnerSource ambassador?

rrrutledge:
> What are the problems that an InnerSource ambassador would address that a central group (e.g. ISPO) cannot easily achieve?

* Eyes, ears, and voice on the ground for the ISPO and supporting its goals in their particular area of the organization.
* Report back to advise the ISPO on what is working and what is not so ISPO is better informed on future actions it takes.

spier:
@rrrutledge @dellagustin-sap based on your input I asked ChatGPT for help, also using some of the info I shared background about the hackathon ambassador idea that I shared above. See #740 for the result.

Also I asked ChatGTP the following:

> What are possible challenges that you see with the described approach of an InnerSource Champion? What would be alternatives?

The answers sounds reasonable. Curious to hear what you think.

### **Challenges**

1. **Workload Balancing**  
   - Ambassadors are expected to perform this role alongside their regular responsibilities, which can lead to overwork or reduced effectiveness in either role.  
   - Without proper support or recognition, ambassadors may become disengaged.

2. **Lack of Authority**  
   - Ambassadors may lack the formal authority to drive change within their departments, making it difficult for them to influence team practices or priorities.

3. **Inconsistent Engagement**  
[...]

PR #740, opened 2024-12-21
[Pattern Draft] InnerSource Ambassador

spier:
Implements #738 

I used the following prompt in ChatGPT as the starting point.

By writing this pattern I would also like to learn how to write better prompts for ChatGPT.
We could then use this prompt to especially help first time pattern authors to create a quick draft, and then iterate from there.

```
Describe the concept of  "InnerSource Ambassadors" in the format of an InnerSource Pattern. Use the template for an InnerSource Pattern provided below.

# The concept:

A large  organization consist of multiple different departments, which are all developing various software components. The organization has decided to invested in InnerSource, to improve collaboration between these departments. An InnerSource Program Office (ISPO) has been established, to drive InnerSource best practices into the whole organization.

However the people working in the ISPO can obviously not be everywhere at the same time. i.e. they are not involved in all departments, regions, and teams of the company. Therefore the ISPO team cannot know how InnerSource is being implemented in the different areas of the company, what challenges the different areas have, what support they need, etc.

Therefore the ISPO decides to look for local InnerSource Ambassadors:

volunteer people drawn from across the organization that are assigned to do a part of supporting InnerSource in their area and across their organization (alongside their regular role in that part of the organization)

Eyes, ears, and voice on the ground for the ISPO and supporting its goals in their particular area of the organization.

Report back to advise the ISPO on what is working and what is not so ISPO is better informed on future actions it takes.

# Template

## Title

Short Title Here

## Patlet

A two sentence description of the problem and solution.
The first sentence describes the problem you are looking to solve, while the second sentence describes the solution.
The first sentence of the Patlet should not mention anything about the solution.

Readers may quickly review dozens of these patlets to discover and browse the larger library of patterns.

## Problem

What is the problem - crisp definition of the problem.

Short description, usually not more than a couple sentences, that describes what the issues and challenges are.

Be careful not to morph into information found in other sections below.

## Story (optional)

Sometimes there is a story that helps people understand the pattern better.

## Context

Where does the problem exist?

What are the pre-conditions?

**Unchangeable** before the solution goes into place.

The content here is often tied to the applicability of the pattern for other readers: "Do I have this same particular situation?"

Use bullet points for this section, so that readers can quickly check if this applies to them.

## Forces

What makes the problem difficult?

What are the trade-offs?

These are constraints that **can be changed** at a cost.

The solution might change one or more of these forces in order to solve the problem, while also in-turn changing the context.

Use bullet points for this section.

## Sketch (optional)

visual illustration

## Solutions

Verified resolutions and possible resolutions to the problem.

## Resulting Context

What is the situation after the problem has been solved?

The original context is changed indirectly by way of the solution.

Often this section can include discussion of the next possible Patterns/problems introduced.

This section can be short in content - the solution may not introduce new problems or change much context.

## Rationale (optional)

Explains why this is the right solution; using totally different words WHY this solution balances these forces and this context to solve this problem.

Can expand on what-if's or theories.

## Known Instances (optional)

Where has this been seen before?

Helps to reinforce that this is a REAL pattern and that you match the context.

May mention:

* A particular business

* Anonymized instances ex: "3 companies have proven that this is a good solution" or "A large financial services org...".

## Status (optional until merging)

General pattern status is stored in GitHub's Label tagging - see any pull request.

Note that this GitHub label tagging becomes less visible once the pattern is finalized and merged, so having some information in this field is helpful.

You might store other related info here, such as review history: "Three of us reviewed this on 2/5/17 and it needs John's expertise before it can go further."

## Author(s) (optional)

Often, this is yourself.

If you need to, find someone in the InnerSource Commons to be the nominal author (As Told To).

Could also be no-one if you do not want to take on authorship (common with a donut looking for a solution).

## Acknowledgments (optional)

Include those who assisted in helping with this pattern - both for attribution and for possible future follow up.

Though optional, most patterns should list who helped in their creation.

## Alias (optional)

If this pattern is also known under a different name than what is listed unter **Title**, please list those alternative titles here.

e.g. if the pattern is named after the problem it solves, a helpful alias might be one that describes the solution that is applied.
```


patterns/1-initial/innersource-ambassador.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Sketch (optional)
+
+*A diagram showing the ISPO at the center, with ambassadors positioned in various departments acting as bidirectional conduits of information and influence.*
+
+## Solutions
+
+- Identify and recruit volunteer ambassadors from across the organization who are enthusiastic about InnerSource. They support InnerSource goals while maintaining their primary organizational roles.
+- Train ambassadors on InnerSource principles, tools, and the goals of the ISPO.
+- Establish clear expectations for ambassadors, including acting as a liaison, promoting InnerSource practices, and providing feedback to the ISPO.
```
spier, line 40: This here might be the most valuable part of the Solution, maybe of the whole pattern.

Can we describe specific tasks that we expect the ambassador to perform?

Some ideas:

- consult local engineering managers when they are looking to make one of their projects ready for InnerSource
- help teams in the discovery phase when starting a major new implementation task - is there any existing project in the org that the team could reuse and contribute to, instead of building something entirely new? (this might require collaboration with other roles like Architects / Principals)

michael-basil:
This is a beautiful pattern and was what we were aiming for during my time at SAP.  In my upcoming role I could see utilizing or introducing this pattern depending on how things play out.

rrrutledge (approved):

patterns/1-initial/innersource-ambassador.md
```diff
@@ -0,0 +1,73 @@
...
+- Initial
+
+## Authors
+
+- Sebastian Spier
+- ChatGPT :)
+
+## Acknowledgments (optional)
+
+## Alias (optional)
```
spier:
```suggestion

## Acknowledgments

- [Russell R. Rutledge](https://github.com/rrrutledge) for sharing the idea and providing feedback
- [Guilherme Dellagustin](https://github.com/dellagustin-sap) for sharing a similar concept used for open source adoption 

## Alias
```

patterns/1-initial/innersource-ambassador.md
spier: Will leave this comment open for reference.

sallydeering1:
@spier This does a great job at capturing all of the elements for IS ambassadors.  To answer your questions, yes it's a good description of what we are doing.  Some additions I would suggest are: a) where you say "are enthusiastic about IS", I would add "and ideally have experience using open source practices". b) add info on 'what's in it for the ambassadors'. For example, it provides opportunity and topic for ambassadors to reach out and engage beyond their immediate teams. it provides opportunity for learning, enhancing influence, problem-solving, and communication skills, opportunity to share their expertise and passion in a structured way, and to network with like-minded enthusiasts.  re: whether Capital One is an org that has implemented the pattern:  I would not list Capital One yet.  What is described in the pattern is our intent, but we are not yet an example of one that has fully implemented the pattern.

PR #741, opened 2024-12-28
Create dependabot.yml

taku333:
# Summary
A mechanism has been added to automatically suggest upgrades to Official Actions and Generic Actions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


spier:
Thanks @taku333!

I have used dependabot before but I don't know what the effect of activating `package-ecosystem: "github-actions"` would be.

Will this create automatic Pull Requests, once a new version of the GitHub Actions is available, or how will this work?

spier:
I found the info that I was looking for here:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

So I have a rough idea of how this will work.

I say let's just try it out. :)

Thank you for your contribution @taku333!

spier:
This has certainly worked, as some PRs were created immediately :)

Screenshot 2024-12-28 at 09 18 08

PR #746, opened 2024-12-28
Mark link as non-reachable in the Unified Source Code Inventory pattern

spier:


spier:
@dterol23 I have marked this link as no longer reachable, so that our automatic link checks don't fail anymore.

However if you somehow get your hands on this content again please let us know and we can fix the link.

PR #747, opened 2024-12-28
Fixing GHA generate-mindmap

spier:


spier:
Once this GHA runs on `main` it throws an error:

```
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: 
remote: - Changes must be made through a pull request.    
```

See for example:
https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/11692137129/job/32560820469

So while this PR won't fix the root cause, it might make the failing runs of the generate-mindmap GHA stop, at least for now :)

PR #748, opened 2024-12-29
Actions version was modified only for the merger.

taku333:
# Summary
eriodically, version upgrade requests come in via dependabot.yml.
Since there is a major version tag, I submitted a request thinking that I would not have to modify it so often.

https://github.com/lycheeverse/lychee-action/tree/v2


taku333:
Close once because it contains unrelated commit logs.

Issue #788, opened 2025-03-10
Broken link in InnerSource Portal pattern


rmarting:
The content of the InnerSource Portal Pattern includes a broken link. The reference is about the Amercian Arilines known instance, basically there is a link to an Internal InnerSource Marketplace at [2020-10-30-innersource/](https://tech.aa.com/2020-10-30-innersource/) reference. That reference is out.

As it could be hard to find the right place, a quick fix could be remove that link.

spier:
Thanks for reporting the issue @rmarting!

We have an automated link check that runs periodically. There it looks like the link broke somewhat recently.
Possibly American Airlines took down their whole tech blog, which would be sad of course.

I will check in our community if we have somebody from American Airlines around that can help us figure out what to do here.

And yes, as you said, the easiest way would be just to remove that link.

rmarting:
Just in case @spier, I opened this PR to fix it (and open to add the right reference for AA). HTH

spier:
Fixed. Thanks for reporting this!

PR #790, opened 2025-03-11
Fix broken link

rmarting:
Small PR to remove a broken link in the [InnerSource Portal pattern](https://patterns.innersourcecommons.org/p/innersource-portal) reported in the issue #788 . 

If there is another good reference to replace the broken link, please, add here as comment, or directly in the PR. Thanks.


spier:
I opted to fix the link via waybackmachine, rather than removing it.

Hope it was ok that I repurposed your PR for this :)

PR #791, opened 2025-03-12
[Pattern Draft] Incentives and Disincentives for InnerSource Contribution

gruetter:
This PR adds a new pattern based on the discussion about incentives and disincentives for InnerSource contribution on the InnerSource Gathering Berlin 2025.


spier (commented):
Thank you for typing this up so quickly @gruetter!

For this idea, I am wondering what would qualify as "adoption" of the pattern? My worry is that this might never make it out of the Initial stage i.e. we would not find known instances for it.

Possibly this type of content might be better suited for other forms of information sharing about InnerSource. Maybe the LearningPath, maybe a bog post, ... maybe something else? Could just be a core question about what is a pattern and what is not?

I am really not sure. Curious to see what others here think about it.

However I am also happy to just try out what happens with this i.e. merge this into the repo, and then see if anybody in the community takes up the idea. Doesn't do any harm either.

gruetter:
Good point, @spier. Here's an idea: count the pattern as adopted, if an org publishes such a list for their trusted committers and contributors. Nothing else contes to mind. That said, I can imagine sharing this in another form. Not sure if this (our patterns repo) is the place where people would (actively) look for this kind of information.

spier:
@gruetter I will just merge this as is. Then we can see if anybody picks this up.

Definitely feel free to use this content for other purposes if you like.

PR #798, opened 2025-03-22
Add Bosch to innersource-ambassador.md

spier:
Adding Bosch as a Known Instance to the InnerSource Ambassador pattern.

Would be great to add a bit more information like since when the concept is in use, how many people fill this role, how people are selected, etc.. Just 1-3 sentences extra description would be enough, as it helps us to explain that every org may implement the pattern slightly differently.


spier:
@gruetter would be great if you could help us find somebody who can add more details about the implementation of this pattern at Bosch.

spier:
> @gruetter would be great if you could help us find somebody who can add more details about the implementation of this pattern at Bosch.

@gruetter any look finding a colleague at Bosch who can help on this PR?

spier:
@gruetter one last reminder on this one ;)

If you cannot find somebody else from Bosch to contribute to this PR, no worries.
We can then just merge this as is.

spier:
It took too long to get further feedback on this. Therefore i am merging as is.

PR #816, opened 2025-05-07
chore: fix typos in pattern system and board reports

pstoeckle:
- used [`typos`](https://github.com/crate-ci/typos) to find and fix some typos


spier (approved):
LGTM

spier:
Thank you for these fixes @pstoeckle, and great to see that you found our repo (I suspect due to the pattern from Siemens that we merged today?).

We already use a spell and style checker (through a [vale GitHub Action](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/.github/workflows/vale.yml)), however we only apply this to the files in `patterns/2-structured/`, which is what gets published in our book https://patterns.innersourcecommons.org.

So in a way I am glad to see that the `typos` tool only found issues in other folders :)

I will get your fixes merged immediately.

pstoeckle:
> Thank you for these fixes @pstoeckle, and great to see that you found our repo (I suspect due to the pattern from Siemens that we merged today?).

Yes 😉

PR #818, opened 2025-05-09
Japanese translation: Convert the toc into sequential text

rashikawat:
The order of the table of contents was all over the place, so I fixed it so that I could read them in order. 
I also categorized each book by the main headings that were in the mind map.


spier:
Thank you @rashikawat for trying to improve the discoverability of the Japanese pattern translations.

The TOC is actually autogenerated programmatically. So if we want to change the order in there, we would have to change some scripts etc.

Interesting idea to create a different pattern grouping. Maybe we could express that in a markmap mindmap as well, if it should not work in the TOC directly.

@yuhattor could you help with this PR, as I cannot review the Japanese content that well ;)

FYI I have limited internet access at the moment, so further input from me will take some weeks.

rashikawat:
@spier Thanks for your comment!

Yes, what I wanted to do was to change the entire table of contents accordingly, since the current mind map is displayed as images and it is not possible to follow links to each pattern. 
Also, the mind map has not yet been translated into Japanese.

So, rather than taking an approach of dynamically creating the list with a script, I decided to fix it and make it a static order for reading.

yuhattor:
Thanks @rashikawat for improving the readability of the Japanese TOC! This manual organization makes much more sense for sequential reading.

I've added a comment to the TOC section to document this change for future maintainers:

`<!-- Custom TOC ordering for enhanced readability - maintain when adding new patterns -->`

This will help ensure that when new patterns are added, contributors know to update the manual ordering you've created. Great work on making the Japanese patterns more discoverable and easier to follow!

yuhattor (approved):
LGTM!

spier:
@yuhattor @rashikawat thank you for your work on the Japanese book.

I took a look at the Japanese book
https://patterns.innersourcecommons.org/ja

Your changes to the TOC may had some side-effects that were not intended.

Please check the points below, and decide if you want to keep them as is or fix some of them.

## (1) Breaking changes to the links

Your change means a breaking change to the Japanese pattern URLs. e.g.
old: https://patterns.innersourcecommons.org/ja/p/start-as-experiment
new: https://patterns.innersourcecommons.org/ja/karasuru/start-as-experiment

Btw the "p" in the URL was configured by [this entry](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/818/files#diff-d9a06302d052f2165d41a0d847f17075ae5162bf2732a2b8e5935395c40822bdL20) in the TOC.

## (2) Really long URLs

gitbook uses the title of sections in the TOC to form the URL.

For some patterns you have these really long URLs now. e.g.

https://patterns.innersourcecommons.org/ja/readmemdcontributingmdkomyunikshontsrunadosuruniha/base-documentation

## (3) New Japanese patterns are not automatically added to the book 

By making the TOC static, new patterns translated to Japanese will not be added to the book automatically.
i.e. for each pattern you will have to remember to add them manually to the TOC

yuhattor:
Good catch!
I fixed it now

spier:
> Good catch! I fixed it now

@yuhattor come again please. Which of the did issues above did you fix, and where/how did you fix them?

yuhattor:
I fixed these issues.

1.  Breaking changes to the links
2. Really long URLs

I acknowledge that 3 is still a problem. I added the comment. Let's add new patterns manually for now and revisit in few month! What do you say?
>

spier:
@yuhattor the links in gitbook now look like this though, which is also a change:
https://patterns.innersourcecommons.org/ja/start-as-experiment

i.e. the "/p/" in the URL is still gone.

You need to add [this entry back](https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/818/files#diff-d9a06302d052f2165d41a0d847f17075ae5162bf2732a2b8e5935395c40822bdL20) into the toc to get the "/p/" in the URL.

yuhattor:
I didn't notice that. Sorry about that.

I now fixed that
https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/832

Issue #805, opened 2025-03-27
Formatting of newlines in gitbook


spier:
gitbook is the tool that we use to host https://patterns.innersourcecommons.org.

gitbook has (recently) changed the way that newline characters (`\n`) in our markdown files are rendered.

Previously a single newline character did not generate a line break.
That meant that we could format our pattern files like below to allow for easier line-based commenting when reviewing patterns.

```
There's a central project that everyone depends on.
What a great candidate for InnerSource!
Unfortunately, the project has grown organically, with various contributions and additions slapped on haphazardly.
```

Recently gitbook made change in this rendering logic. Now such text in markdown is rendered exactly as shown i.e. each newline in the markdown also leads to a line break in the rendered HTML. This does not look great. See screenshot below.

![Image](https://github.com/user-attachments/assets/b7d12a3f-25ba-459e-a569-93d1288bffe5)

dellagustin-sap:
Maybe this was a bug introduced in Gitbook, this seems like a pretty noticeable change, but I could not find a record of it at https://gitbook.com/docs/changelog with a quick inspection.

dellagustin-sap:
@spier have you tried reaching out to their support?
https://gitbook.com/docs/help-center/further-help/how-do-i-contact-support

spier:
Yes. Their Support says that nothing was changed in their side. I am reasonably sure that that is not true but I cannot really prove it.

Even proving it would not help me. Rather we have to find a way to make the new rendering logic of gitbook work for us. Or switch to an entirely different solution but they would be a lot more work.

dellagustin-sap:
Not sure why I'm going through this rabbit hole instead of working on my to-do list, but hey sometimes that happens 😄 .
Their documentation on Markdown is quite meager - https://gitbook.com/docs/creating-content/formatting/markdown, but they do refer to CommonMark, os they are not reinventing the wheel. 
CommonMark states (https://spec.commonmark.org/0.31.2/#soft-line-breaks):
> A regular line ending (not in a code span or HTML tag) that is not preceded by two or more spaces or a backslash is parsed as a [softbreak](https://spec.commonmark.org/0.31.2/#softbreak). (A soft line break may be rendered in HTML either as a [line ending](https://spec.commonmark.org/0.31.2/#line-ending) or as a space. The result will be the same in browsers. In the examples here, a [line ending](https://spec.commonmark.org/0.31.2/#line-ending) will be used.)
> ...
[...]

PR #823, opened 2025-06-04
Update maturity-model.md - minor formatting change

dellagustin-sap:
Moved a split sentence into a single line, trying to solve a rendering issue on GitBook. "development" and "forge" are rendered without a space between the two words.

screenshot of the Forces section of the Maturity Model pattern, point out that the development and forge words are rendered as one

### DIFF
--- a/patterns/2-structured/maturity-model.md
+++ b/patterns/2-structured/maturity-model.md
@@ -33,8 +33,7 @@ InnerSource project differs depending on each team.
 Teams sharing InnerSource learnings run into misunderstandings as they are not
 aware of their respective level of InnerSource adoption.
 
-Teams believe that "it's all about migrating to a shared software development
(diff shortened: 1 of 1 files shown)


spier (approved):

spier:
Thanks for the fix @dellagustin-sap. This should go live shortly. Could you check if it looks right now?

Btw this is likely related to #805

dellagustin-sap:
Hi @spier , I rechecked now, either the fix did not work, or it did not go live.
Is there a way to check if it went live, besides checking the content?

spier:
The sync to gitbook failed. I just updated it manually.

Now the changes are live here:
https://patterns.innersourcecommons.org/p/maturity-model

spier:
> Is there a way to check if it went live, besides checking the content?

No, unfortunately not. You have to check the content itself to know if the latest commits went live.
As an admin on our gitbook space I can debug this better but that info is not available publicly.

dellagustin-sap:
@spier I checked and it fix worked. Thanks for the manual publishing.

PR #826, opened 2025-06-22
feat: Weaving the Garden Grid pattern

michael-basil:
Hello @spier, @rrrutledge, @Trueblueaddie, @claredillon, @dellagustin, @tsadler1988  and others!

I have been reflecting on my experiences in various corporations over the years ...

I feel that this pattern is important to get clear on.

I have been reading _Fearless Change_ by Mary Lynn Manns, PhD recently so I started with that format to get things into a first frame.

If it pulls you in, I would love to get some feedback!

⛩️🌿

### DIFF
--- a/patterns/1-initial/garden-grid.md
+++ b/patterns/1-initial/garden-grid.md
@@ -1,0 +1,84 @@
+# 🌿 Weaving the Garden Grid
+
+## Also Known As
+- *Circles Before Systems*  
+- *Middle-Out Mycelium*  
+- *Feedback as Fertilizer*  
+
+---
+
+## Context
+
+In large, highly regulated organizations undergoing agile or digital transformation, distinct groups such as development, compliance, security, and architecture operate with different priorities, languages, and timelines. Each tends its own “plot,” guided by its own definitions of care, quality, and risk.
+
+The system aspires to become more innovative, adaptive, and aligned — but is held back by structural inertia, missing feedback loops, and tribal mistrust between departments.
+
+---
+
+## Problem
+
+Transformation efforts stall when teams attempt to grow in isolation. Developers innovate without understanding compliance; compliance defends policy without seeing current delivery realities. Architects design elegant futures without grounding in the messy present. Communication becomes brittle, misaligned, and deeply siloed — resulting in plans that don’t root, policies that don’t land, and humans who burn out.
+
+---
+
+## Forces
+
+- 📜 **Compliance mandates** create rigidity in documentation and approval.  
+- 🧠 **Security demands** shift focus toward control and risk elimination.  
+- 🏃 **Agile teams** push for velocity and autonomy.  
+- 🧱 **Architects** seek elegant future states across messy current states.  
+- 🌍 **Global scale** amplifies inconsistencies across business units.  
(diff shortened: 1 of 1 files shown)


patterns/1-initial/garden-grid.md
```diff
@@ -0,0 +1,84 @@
+# 🌿 Weaving the Garden Grid
```
spier:
```suggestion
## Title

🌿 Weaving the Garden Grid
```

patterns/1-initial/garden-grid.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Also Known As
```
spier, line 5:
```suggestion
## Also Known As

```

Fixing markdownlint

spier (commented):
Thanks for sharing this @michael-basil.

I left two inline comments to fix technicalities.

Unfortunately I likely won't have much more time this week to review this. But maybe one of the other folks that you tagged does have time?

Just two notes related to GitHub, that makes working with these PRs easier. No need to fix this for this PR, just ideas for next time:

- when you open a PR from your main branch, then I cannot commit changes back to that (branch protection)
- when you open a PR from an org repositority (rather than a user repository), github does not allow me to commit changes back to your branch at all / don't know why

michael-basil:
Hi @spier!

Thanks for the first look.  Still working through the draft, but the content is in good enough shape for more people to weigh in as they would have availability and interest.

I don't believe there is branch protection enabled on this forked repository ... so I added you as an outside contributor to the repo.  Perhaps that will work if there are quick changes you'd want to make?

Looking forward to collaborating again!

⛩️🌿

dellagustin-sap:
> Just two notes related to GitHub, that makes working with these PRs easier. No need to fix this for this PR, just ideas for next time:
> 
> * when you open a PR from your main branch, then I cannot commit changes back to that (branch protection)
> * when you open a PR from an org repositority (rather than a user repository), github does not allow me to commit changes back to your branch at all / don't know why

@spier , good to know, thanks for sharing.

dellagustin-sap:
@michael-basil , I'm still reviewing the content, but a "technical" comment: This is something I think you will appreciate, have you heard about the concept of "Semantic Line Breaks"?
https://sembr.org/
I think this is is use in other parts of ISC, but I don't recall where I saw it first. It is not part of the contribution guidelines of the Patterns, but I though it would be interesting to share. However, we recently observed that GitBook has changed and does not support this practice very well, see https://github.com/InnerSourceCommons/InnerSourcePatterns/issues/805.

dellagustin-sap:
@michael-basil , I did my first read, but would need to dig deeper to comment more.
One general comment, it would be good to make the connection of this pattern to InnerSource more explicit. I can infer some connections, but I think such connections should be clear from the text.

michael-basil:
Thank you @dellagustin-sap and @spier for weighing in on this.

I have reflected on this proposed pattern with a couple of people in Mindset Dojo and tested for interest from former colleagues at SAP.  

Based on those discussions and the commentary here, it does not seem to have enough motivation to carry forward at present.

Therefore I am closing the pull request and will see what InnerSource synerges emerge in time.

Thank you for the engagement and consideration.

🙏
⛩️🌿

PR #838, opened 2025-09-06
Fix link to DBSystel InnerSource License

spier:
Noticed a broken link in [this check](https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/17488150853).


lpanni:
I noticed that the link on the website is still broken (the website shows 'Last updated 6 months ago') so this change is not deployed there.

spier:
Great catch @lpanni!

Sometimes the sync from github to our hosting service (gitbook) is a bit flaky.

For the English site this is fixed now.
Now I have to do the same for all translations :(

spier:
Now it is fixed for all 😓

PR #571, opened 2023-08-09
Trying new GHA for taking a screenshot

spier:
Experiments related to #570.

PR #849, opened 2025-09-15
Mindmap screenshot with pageres-cli

### LINKED ISSUE #570: Fix generate-mindmap GHA workflow
When running our [generate-mindmap](https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/.github/workflows/generate-mindmap.yml) GitHub Action workflow, we get these two warnings:  

> The following actions uses node12 which is deprecated and will be forced to run on node16: swinton/screenshot-website@v1.x. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/Show

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Currently the screenshot of our mindmap still gets generated correctly by the generate-mindmap workflow. However at some point in the not too distant future this will stop working.

Looks like both warnings are related to the `swinton/screenshot-website` GHA that we are using for generating a screenshot of the HTML generated by markmap. However that GHA does not seem to be maintained anymore, although we are trying to reach the maintainer [here](https://github.com/swinton/screenshot-website/issues/37).

What options do we have?

1) switch to another GHA that can create a screenshot from HTML
2) help fix `swinton/screenshot-website` and create a new release that we (and others) can use

Option (1) might be the quickest, although (2) would also be neat in the spirit of open source :)

spier:
Fixes #570. (or at least tries to ... again ;))

I am trying again to fix the GHA workflow that takes a screenshot of the mindmap that is rendered in our online book.
Previous failed tries were #571 and #775.

Found this here that can take screenshots on the CLI:
https://github.com/sindresorhus/pageres-cli

Could not find a GHA that wraps this but it was easy enough to express this in a workflow myself.

TODO:

- [x] once this PR is merged, remember to delete all the old branches of the closed PRs referenced above
- [ ] possibly post an update on the old GHA that I used, to point users to pageres-cli


spier:
The screenshot generation works now.

However the "Commit Changes" step does not work, as multiple matrix builds are trying changing the state of the repository in parallel. The git-auto-commit-action GHA explicitly [mentions](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#limitations--gotchas) this limitation.

One way to solve this would be to upload the screenshots as artefacts in the matrix build, then at the end have a single job that downloads all artefacts and commits them. See https://github.com/stefanzweifel/git-auto-commit-action/discussions/290 and https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#using-an-output-to-define-two-matrices

spier:
I tried a configuration where only one job of this workflow can run at the same time. i.e. switching off concurrency for this workflow entirely.

However even then [it returns an error](https://github.com/InnerSourceCommons/InnerSourcePatterns/actions/runs/18016950140/job/51264268966):
"Updates were rejected because the remote contains work that you do not have locally." I don't get it :(

spier:
Since https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/849/commits/f4ad3fbcf9bda95585d6b3096bc95c43ce0559d2 the execution is working.

All jobs of the matrix build are running sequentially now (aside: the whole `concurrency` config could be removed now)

Each of these jobs takes 30sec though, so in total the workflow takes 2 minutes, which is quite long.

I am considering to put it all in a single script that loops over all folders below "pattern-categorization" and calls the required clis to generate the markmap, take the screenshot, etc.

spier:
YES, finally!

Instead of using a GHA matrix build, I ended up with this approach:
- Created a single script that loops over all folders below "pattern-categorization".
- Then calls the required clis to generate the markmap, take the screenshot, etc.
- At the end there is a single commit, pushing all changes to the repo (this way there cannot be any issues with the repo being out of sync with what is on the branch)

PR #867, opened 2025-11-14
Adding research information to Career Booster pattern

spier:
Klaas-Jan Stol, Mario Schaarschmidt, Lorraine Morgan: Does adopting inner source increase job satisfaction? A social capital perspective using a mixed-methods approach. [Research paper](https://www.sciencedirect.com/science/article/pii/S0963868724000015).

### DIFF
--- a/patterns/1-initial/innersource-as-career-booster.md
+++ b/patterns/1-initial/innersource-as-career-booster.md
@@ -55,6 +55,24 @@ The organization sees better collaboration and higher retention among high-perfo
 
 This creates a reinforcing loop where career incentives and organizational goals align more closely.
 
+### Consequences
+
+#### Positive:
+
+* Broader network across the organization → more opportunities and visibility.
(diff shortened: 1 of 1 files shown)


spier:
@user-group I believe your research is highly relevant for this pattern.

Please note that this is a draft pattern i.e. no org has confirmed it yet.

So far our pattern is called "InnerSource as a Career Booster", however I feel that job satisfaction and success in the job a pretty closely linked. So I hope it is not unrealistic to try to integrate your research findings into this pattern.

I tried to add parts of your research to this pattern already.

However I would be curious to hear what you think about this pattern in general, and which other highlights from your research paper you would suggest to add to it.

spier:
I did not receive feedback for this changes.

Still I am opting to merge them as is, as adding the link from this pattern to the research page should definitely be worthwhile. Also the pattern is still maturity Initial, so before it makes its way to the book it can still be reviewed/improved.

PR #873, opened 2025-11-25
Update dependabot.yml

lenucksi:
Keeps producing ruby security warnings that seem to be outdated versions https://github.com/InnerSourceCommons/InnerSourcePatterns/security/dependabot/40 for the book scripts.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "weekly"


spier:
Hi @lenucksi. Great to see you here again!

Is this the goal of this change to run the dependency checks for the ruby dependencies less frequently, and with that generate fewer alerts? If so, we could even bring this down to **monthly**, to reduce the alerts even further.

Also we can definitely try if the book generation also works with commonmarker >= 0.23.10, if that fixes the alerts.
I have not touched this script in a long time.

spier:
Just FYI, I fixed the dependabot alerts related to `generate_toc.rb` in  https://github.com/InnerSourceCommons/InnerSourcePatterns/pull/879.

Still happy to merge this PR here as well, if it helps to reduce the alerts further.

PR #879, opened 2025-11-25
Upgrade to commonmarker 0.23.10 to fix the dependabot alerts

spier:
Upgrade to commonmarker 0.23.10 to fix the various dependabot alerts


spier:
I verified that after the upgrade the script still generates the same TOC for the English book as before.
Good enough as a quick test.

Merging...

PR #906, opened 2026-05-05
Fix missing patlets

rrrutledge:
- Fix # Patlet heading level to ## Patlet in 3 files (balancing-openness-and-security, code-consumers, explicit-shared-ownership)
- Write patlets for 14 patterns that had TBD placeholder content
- All patlets derived from each pattern's existing Problem/Context/Solution sections
- The purpose of doing this work is to provide a sort of "index" for AI to determine what a particular pattern is about without having to take its full text into its context.


jeffabailey (approved):

spier (approved):
I did not have time to sanity check whether the Patlets match the pattern's content. However as @jeffabailey already approved the PR, I assume these are fine ;)

I added the new Patlets to the main README as well. There you can also see that there are a couple more patterns left without a Patlet, if you care to fix even more in further PRs.

As an aside: I struggle a bit with some of these draft patterns, as some are rather weak content-wise, and we should at least consider to deprecate/delete them. However having Patlets for them does not hurt in either case.

Thank you for doing this important cleanup work! Merging shortly ... 🥳

PR #926, opened 2026-08-11
Add GitLab as Known Instance and draft Paired Onboarding Sprint pattern

rrrutledge:
## Summary

From Raimund Hook's InnerSource Commons talk ["Break the contribution wall"](https://www.youtube.com/watch?v=d3igPlTr4Ag) (April 2026). GitLab's DevRel Engineering team (formerly Contributor Success) runs a contributor-growth program reaching "hundreds of users a week."

- **Incentive Mechanisms for Voluntary Contributions** — adds GitLab as a Known Instance: [contributors.gitlab.com](https://contributors.gitlab.com) gamifies contribution across code, code review, documentation, and translations with points, levels, and leaderboards; points unlock exclusive contributor-only swag - replaces "TBD" Known Instances
- **Internal Developer Platform** — adds GitLab as a Known Instance: GitLab containerized their GitLab Development Kit (GDK), cutting new-contributor environment setup from roughly two hours (frequently failing) to under five minutes
- **Paired Onboarding Sprint** (new pattern, filed at maturity level 1/Initial) — GitLab's "Co-create" program pairs a host-team engineer with a high-value potential contributor for a dedicated week to make their first real contribution together. No existing pattern covers this structured 1:1 onboarding model; one known instance is exactly what Initial-level requires per `meta/contributor-handbook.md`.

## Test plan

- [ ] Verify all YouTube links resolve to the correct talk
- [ ] Confirm the two Known Instance entries match the surrounding formatting style
- [ ] Confirm Paired Onboarding Sprint follows the pattern template and reads clearly for a first-time reader

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #927, opened 2026-08-12
Fix link checker false positives on root-relative asset links

rrrutledge:
## Problem

The link checker workflow only lints files changed in a PR, specifically to avoid failing PRs on pre-existing issues unrelated to their content — the workflow's own top comment says so.
But it still checks every link in a changed file, not just the diff, so any pre-existing root-relative link (e.g. `/assets/img/foo.png`) in a touched file trips it, since lychee isn't told the repo root and can't resolve a path starting with `/`.

This surfaced in #926, which failed the check on a pre-existing image link in `patterns/1-initial/internal-developer-platform.md` that the PR never touched — it only added a "Known Instances" bullet elsewhere in the file.

## Fix

Pass `--root-dir .` to lychee so root-relative links resolve against the repo root instead of failing outright.

### DIFF
--- a/.github/workflows/link-checker-prs.yml
+++ b/.github/workflows/link-checker-prs.yml
@@ -39,7 +39,7 @@ jobs:
         name: Link Checker
         uses: lycheeverse/lychee-action@v2
         with:
-          args: --verbose --no-progress --cache --max-cache-age 1d $MARKDOWN_FILES
+          args: --verbose --no-progress --cache --max-cache-age 1d --root-dir . $MARKDOWN_FILES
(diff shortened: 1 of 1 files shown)


spier (approved):

spier:
@rrrutledge thank you for fixing this issue in the GHA.

I suspect we should apply the same fix to a similar GHA that runs once per week:
https://github.com/InnerSourceCommons/InnerSourcePatterns/blob/main/.github/workflows/link-checker.yml#L30?

I am doing that now.

